PR #110, opened 2022-07-25
Add PostgreSQL, migrations

### LINKED ISSUE #76: Projekto pradinė konfigūracija

adp-atea:


pyproject.toml
```diff
@@ -6,16 +6,19 @@ authors = ["IVPK <atviriduomenys@ivpk.lt>"]
...
 django-sekizai = "^3.0.1"
 django-filer = "^2.2.1"
 djangocms-text-ckeditor = "^5.0.1"
 djangocms-blog = "^1.2.3"
 django-compressor = "^4.0"
 django-libsass = "^0.9"
 libsass = "^0.21.0"
-django-sass-processor = "^1.2"
+django-sass-processor = "^1.2.1"
+mysqlclient = "^2.1.1"
```
sirex: As I understand, MySQL is no longer needed?

pyproject.toml
```diff
@@ -6,16 +6,19 @@ authors = ["IVPK <atviriduomenys@ivpk.lt>"]
...
+django-environ = "^0.9.0"
+psycopg2 = "^2.9.3"
```
sirex, line 22: I would recommend precompiled binaries, that comes with [psycopg2-binary](https://pypi.org/project/psycopg2-binary/) package.

vitrina/classifiers/models.py
```diff
@@ -21,7 +21,7 @@ class Category(models.Model):
     icon = models.CharField(max_length=255, blank=True, null=True)
 
     class Meta:
-        managed = False
+        managed = True
```
sirex, line 24: All models should be `managed = False`, database should be created from a PostgreSQL SQL dump, same thing will be used in production. PostgreSQL database will be created and updated using [PGLOADER](https://pgloader.io/).

So do not create database using Django migrate, create it using PGLOADER. Once we are done with the migration, then we will switch all models to `manage = True` and switch to  Django migrations. Until then, we have to stick with PGLOADER.

vitrina/models.py
```diff
@@ -5,1421 +5,1423 @@
...
+#     modified = models.DateTimeField(blank=True, null=True)
+#     version = models.IntegerField()
+#     rating = models.IntegerField()
+#     dataset = models.ForeignKey(Dataset, models.DO_NOTHING, blank=True, null=True)
+#     user = models.ForeignKey(User, models.DO_NOTHING, blank=True, null=True)
+#     harvested = models.ForeignKey(HarvestingResult, models.DO_NOTHING, blank=True, null=True)
+#
+#     class Meta:
+#         managed = True
+#         db_table = 'user_vote'
```
sirex: Could you please just remove all the commented out code? We don't need comment out code, since all the code and all the changes are available under version control and we can restore everything we need.

Thanks.

vitrina/settings.py
```diff
@@ -135,12 +135,12 @@
...
+        'ENGINE': 'django.db.backends.postgresql',
         'NAME': 'adp-dev',
-        'USER': 'root',
-        'PASSWORD': 'r00t',
-        'HOST': '192.168.2.60',
-        'PORT': '3306'
+        'USER': 'adp',
+        'PASSWORD': 'secret',
+        'HOST': '127.0.0.1',
+        'PORT': '5432',
```
sirex: Could you please change that to [environ.Env.db_url](https://django-environ.readthedocs.io/en/latest/types.html#environ-env-db-url)?

sirex (changes requested):

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+      POSTGRES_PASSWORD: secret
+    volumes:
+      - ./postgres:/var/lib/postgresql/data
+    ports:
+      - 127.0.0.1:5432:5432
+
+  adminer:
+    image: adminer
+    ports:
+      - 0.0.0.0:9000:8080
```
sirex: Maybe change that to

```suggestion
      - 127.0.0.1:9000:8080
```

in order to not expose that to the outside network?

README.rst
```diff
@@ -1,18 +1,31 @@
...
 .. _GitHub Flow: https://docs.github.com/en/get-started/quickstart/github-flow
+
+
+Development environment
+***********************
+
+From project root directory run::
+
+    docker-compose up
+    poetry run python manage.py migrate
```
sirex, line 31: Also, could you add information, how to connect to PostgreSQL from adminer? I am able to connect from DBeaver, but not from adminer. Maybe adminer needs to be linked to postgres service in docker-compose file?

vitrina/classifiers/models.py
sirex: To be more specific, we need to add same MySQL version as in production to docker-compose, then load MySQL SQL schema file into MySQL instance running on Docker, and then run PGLOADER, to sync schema and data to PostgreSQL instance running on Docker.

This way we will be able to fully test migration and switch to Django migrations only when we are running PostgreSQL on production.

We can add Django migrations for specific models on top of  PGLOADER loaded schema by removeing `managed = False` line and then creating migrations:

    ./manage.py makemigrations

And then skipping initial migration:

    ./manage.py migrate --fake-initial

This way, we can stick to the original tables from PGLOADER and add new schema changes on top, using Django migrations system. But this must be done case by case and only, when we need to change existing models.

sirex: Follow up task: https://github.com/atviriduomenys/katalogas/issues/111

sirex (approved):

PR #112, opened 2022-07-27
Added dataset detail view

adp-atea:

PR #113, opened 2022-07-27
80 rinkinių filtrai

### LINKED ISSUE #80: Rinkinių filtrai
Filtrai pagal

- statusą
- organizacija
- kategorija
- žymę
- formatą
- atnaujinimą
- įkėlimo datą

https://data.gov.lt/datasets

adp-atea:
- [x] Add tests (preferably finish https://github.com/atviriduomenys/katalogas/issues/65 before this).
- [ ] Remove translation files? I think, it is not related to this task and is some kind of artifact left after experimentation with translation strings?
- [x] Multiple filters can be set, as it is now in data.gov.lt.
- [x] Active filter should be highlighted.
- [x] Filters options should be reduced according to current filters set.
- [x] Filters options should show number of datasets.
- [x] datepicker is no longer used and should be removed?
- [ ] External JS/CSS assets should be managed be an asset manager and should not be added to repository.


locale/lt/LC_MESSAGES/django.po
```diff
@@ -91,50 +92,90 @@ msgstr "O gal duomenų rinkinys neatitinka lūkesčių?"
...
-msgid "general.more"
-msgstr ""
-
-#: vitrina/datasets/templates/vitrina/datasets/list.html:249
-msgid "general.less"
+#: vitrina/datasets/templates/vitrina/datasets/list.html:241
+msgid "Suskleisti"
 msgstr ""
 
 #: vitrina/orgs/apps.py:8
```
sirex: Why this file is here? As I understand, this was generated by https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#localization-how-to-create-language-files?

I suggest, to remove this file for now and to translations separately?

pyproject.toml
```diff
@@ -16,6 +16,7 @@ django-compressor = "^4.0"
 django-libsass = "^0.9"
 libsass = "^0.21.0"
 django-sass-processor = "^1.2"
+mysqlclient = "^2.1.1"
```
sirex: I seen this before, but we don't need `mysqlclient`. Please remove.

vitrina/datasets/templates/vitrina/datasets/filters.html
```diff
@@ -0,0 +1,126 @@
+{% load i18n %}
+
+<nav class="category-panel">
+    <p class="category-panel-heading">
+        <strong>{% translate "Rinkinio būsena" as datasetStatus %}{{ datasetStatus|upper}}</strong>
+    </p>
+    <div class="facet-data_status is-flex">
+        <a href="?status=HAS_DATA" class="category-panel-block" title="Atverti duomenys">
```
sirex: It should be possible to filter by multiple filters, now each filter overrides previous filters. Also, currently active filter is not shown.

I would move all filtering logic out of templates to Python, and in templates, just render prepared data, with all the urls, active filters, etc.

vitrina/datasets/templates/vitrina/datasets/filters.html
```diff
@@ -0,0 +1,126 @@
...
+    <p class="category-panel-heading">
+        <strong>{% translate "Įkėlimo data" as dateTitle %}{{ dateTitle|upper}}</strong>
+    </p>
+    <div class="field">
+        <input class="input datepicker-from" name="date_from" id="from"
+                           placeholder="{% translate "Data nuo" %}">
+    </div>
+    <div class="field">
+        <input class="input datepicker-to" name="date_to" id="to"
+                           placeholder="{% translate "Data iki" %}">
```
sirex: These days, most browsers have a built-in date picker, do I'm not sure if we need a custom JavaScript implementation for that?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#browser_compatibility

vitrina/datasets/views.py
```diff
@@ -14,3 +17,37 @@ class DatasetListView(ListView):
...
+                Q(title__icontains=query) | Q(title_en__icontains=query)
+            )
+        elif date_from and date_to:
+            datasets = datasets.filter(Q(published__gt=date_from) & Q(published__lt=date_to))
+        elif date_from:
+            datasets = datasets.filter(published__gt=date_from)
+        elif date_to:
+            datasets = datasets.filter(published__lt=date_to)
+        elif filter_dict:
+            datasets = datasets.filter(**filter_dict)
```
sirex: It should be possible to add multiple filters.

sirex (changes requested):
.

sirex:
I found, that this pull request, is based on top of https://github.com/atviriduomenys/katalogas/pull/115, please, separate changes correctly, do not create pull requests multiple times, for same changes.

This pull request should only include filter related changes, search related changes should not be here.

tests/test_dataset_filters.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+@pytest.mark.django_db
+def test_dataset_filter_status(app: DjangoTestApp):
+    dataset_has_data = DatasetFactory(status=HAS_DATA)
+    dataset_inventored = DatasetFactory(status=INVENTORED)
+    dataset_metadata = DatasetFactory(status=METADATA)
+    dataset_structure = DatasetStructureFactory()
+    dataset_structure.dataset = dataset_has_data
+    dataset_structure.save()
```
sirex:
```suggestion
    dataset_structure = DatasetStructureFactory(dataset=dataset_has_data)
```

tests/test_dataset_filters.py
```diff
@@ -0,0 +1,273 @@
...
+
+    # without query
+    resp = app.get(reverse('dataset-search-results'))
+    assert resp.status == '200 OK'
```
sirex: No need to test status, it is checked by `app.get`.

vitrina/datasets/views.py
```diff
@@ -1,7 +1,26 @@
...
+HAS_DATA = "HAS_DATA"
+INVENTORED = "INVENTORED"
+HAS_STRUCTURE = "HAS_STRUCTURE"
+METADATA = "METADATA"
+DATASET_STATUSES = {
+    HAS_DATA: _("Atverti duomenys"),
+    INVENTORED: _("Tik inventorintas"),
+    HAS_STRUCTURE: _("Įkelta duomenų struktūra"),
+    METADATA: _("Tik metaduomenys")
+}
```
sirex: See https://github.com/atviriduomenys/katalogas/pull/112#discussion_r951349613

In other words, this should be defined in `models.py` as model status choices using Django provided functionality.

tests/test_dataset_filters.py
```diff
@@ -0,0 +1,273 @@
...
+    assert len(resp.context['object_list']) == 1
+    assert resp.context['object_list'][0].pk == dataset_metadata.pk
+    assert resp.context['selected_status'] == METADATA
+
+    # with status HAS_STRUCTURE
+    resp = app.get("%s?status=%s" % (reverse('dataset-search-results'), HAS_STRUCTURE))
+    assert resp.status == '200 OK'
+    assert len(resp.context['object_list']) == 1
+    assert resp.context['object_list'][0].pk == dataset_has_data.pk
+    assert resp.context['selected_status'] == HAS_STRUCTURE
```
sirex: Move separate test cases into separate test case functions. And there is no need to run tests on every possible value of an attribute.

tests/test_dataset_filters.py
```diff
@@ -0,0 +1,273 @@
...
+
+@pytest.mark.django_db
+def test_dataset_filter_organization(app: DjangoTestApp):
+    organization = OrganizationFactory()
+    dataset_with_organization1 = DatasetFactory()
+    dataset_with_organization2 = DatasetFactory()
+    dataset_with_organization1.organization = organization
+    dataset_with_organization1.save()
+    dataset_with_organization2.organization = organization
+    dataset_with_organization2.save()
```
sirex:
```suggestion
    dataset_with_organization1 = DatasetFactory(organization=organization)
    dataset_with_organization2 = DatasetFactory(organization=organization)
```

tests/test_dataset_filters.py
```diff
@@ -0,0 +1,273 @@
...
+
+    # without query
+    resp = app.get(reverse('dataset-search-results'))
+    assert resp.status == '200 OK'
+    assert len(resp.context['object_list']) == 2
```
sirex: It is much better to test content instead of amount. There can be duplicates in results also order is important in most cases.

vitrina/classifiers/models.py
```diff
@@ -27,6 +28,34 @@ class Meta:
...
+        if include_self:
+            parents.append(self)
+        if self.parent_id:
+            try:
+                parent = Category.objects.get(pk=self.parent_id)
+            except ObjectDoesNotExist:
+                parent = None
+            if parent:
+                parents.extend(parent.get_parents(include_self=True))
+        return parents
```
sirex: Use [django-treebeard's get_ancestors()](https://django-treebeard.readthedocs.io/en/latest/api.html#treebeard.models.Node.get_ancestors) instead.

vitrina/classifiers/models.py
```diff
@@ -27,6 +28,34 @@ class Meta:
...
+
+    def get_children(self, include_self=False):
+        all_children = []
+        children = Category.objects.filter(parent_id=self.pk)
+        if include_self:
+            all_children.append(self)
+        for child in children:
+            all_children.extend(child.get_children(include_self=True))
+        return all_children
```
sirex: Use [django-treebeard's get_children()](https://django-treebeard.readthedocs.io/en/latest/api.html#treebeard.models.Node.get_children) instead.

vitrina/classifiers/models.py
```diff
@@ -27,6 +28,34 @@ class Meta:
...
+
+    def get_family_objects(self):
+        parents = self.get_parents()
+        children = self.get_children()
+        parents.extend(children)
+        return parents
```
sirex:
```suggestion
        yield from self.get_parents()
        yield from self.get_children()
```

vitrina/datasets/templates/vitrina/datasets/list.html
```diff
@@ -1,10 +1,16 @@
...
 {% load i18n %}    
 
 {% block pageTitle %} | {% translate "Duomenų rinkiniai" %}{% endblock %}
 {% block pageOgTitle %} | {% translate "Duomenų rinkiniai" %}{% endblock %}
 
 {% block head %}
+    <script src="{% static 'js/jquery-3.6.0.min.js' %}"></script>
+    <link href="{% static 'css/datepicker.min.css' %}" rel="stylesheet">
+    <script src="{% static 'js/datepicker.min.js' %}"></script>
+    <script src="{% static 'js/datepicker.lt-LT.js' %}"></script>
```
sirex: As I understand datepicker is no longer needed?

vitrina/datasets/views.py
```diff
@@ -14,3 +33,196 @@ class DatasetListView(ListView):
...
+    def filter_by_status(queryset, status):
+        if status == HAS_DATA:
+            return queryset.filter(status=HAS_DATA)
+        elif status == INVENTORED:
+            return queryset.filter(status=INVENTORED)
+        elif status == HAS_STRUCTURE:
+            return queryset.filter(datasetstructure__isnull=False)
+        elif status == METADATA:
+            return queryset.filter(status=METADATA)
+        return queryset
```
sirex: Move this to a model manager or `services.py`.

vitrina/datasets/views.py
```diff
@@ -14,3 +33,196 @@ class DatasetListView(ListView):
...
+            if only_children:
+                family_objects = [obj.pk for obj in selected.get_children()]
+            else:
+                family_objects = [obj.pk for obj in selected.get_family_objects()]
+            related_categories.extend(family_objects)
+        if len(selected_category_objects) > 1:
+            related_categories = [category for category in related_categories
+                                  if related_categories.count(category) == len(selected_category_objects)]
+        related_categories = list(set(related_categories))
+        return related_categories
```
sirex: Move to `services.py`.

sirex (changes requested):

tests/datasets/test_views.py
```diff
@@ -0,0 +1,277 @@
...
+def test_dataset_filter_all(app: DjangoTestApp):
+    organization = OrganizationFactory()
+    category = CategoryFactory()
+    frequency = FrequencyFactory()
+    dataset_with_all_filters = DatasetFactory(status=Dataset.HAS_DATA, tags="tag1, tag2, tag3", published=datetime(2022, 2, 9),
+                                              organization=organization, category=category, frequency=frequency)
+
+    resp = app.get("%s?status=%s&organization=%s&category=%s&tags=tag1&tags=tag2&frequency=%s&"
+                   "date_from=2022-01-01&date_to=2022-02-10" % (reverse("dataset-search-results"), Dataset.HAS_DATA,
+                                                                organization.pk, category.pk, frequency.pk))
```
sirex:
```suggestion
    resp = app.get(reverse("dataset-search-results"), {
        'status': Dataset.HAS_DATA,
        'organization': organization.pk,
        'category': category.pk,
        'tags': 'tag1',
        'tags': 'tag2',
        'frequency': frequency.pk,
        'date_from': '2022-01-01',
        'date_to': '2022-02-10',
    })
```

> Beautiful is better than ugly.
> Readability counts.
> -- https://peps.python.org/pep-0020/

tests/datasets/test_views.py
```diff
@@ -0,0 +1,277 @@
...
+    assert resp.context['selected_date_to'] == "2022-02-10"
+
+
+@pytest.mark.django_db
+def test_dataset_filter_all(app: DjangoTestApp):
+    organization = OrganizationFactory()
+    category = CategoryFactory()
+    frequency = FrequencyFactory()
+    dataset_with_all_filters = DatasetFactory(status=Dataset.HAS_DATA, tags="tag1, tag2, tag3", published=datetime(2022, 2, 9),
+                                              organization=organization, category=category, frequency=frequency)
```
sirex:
```suggestion
    dataset_with_all_filters = DatasetFactory(
        status=Dataset.HAS_DATA,
        tags="tag1, tag2, tag3",
        published=datetime(2022, 2, 9),
        organization=organization,
        category=category,
        frequency=frequency,
    )
```

> Beautiful is better than ugly.
> Readability counts.
> -- https://peps.python.org/pep-0020/

vitrina/datasets/views.py
```diff
@@ -14,3 +20,168 @@ class DatasetListView(ListView):
 class DatasetDetailView(DetailView):
     model = Dataset
     template_name = 'vitrina/datasets/detail.html'
+
+
+class DatasetSearchResultsView(ListView):
```
sirex: Please move this into `DatasetListView`, we don't need separate view for search results, we only have on `DatasetListView` which always shows a list of datasets, optionally filtered.

vitrina/datasets/views.py
```diff
@@ -14,3 +20,168 @@ class DatasetListView(ListView):
...
+        filter_dict.pop('tags', None)
+        filter_dict.pop('category', None)
+        datasets = Dataset.public.order_by('-published')
+
+        if query:
+            datasets = datasets.filter(
+                Q(title__icontains=query) | Q(title_en__icontains=query)
+            )
+        if date_from and date_to:
+            datasets = datasets.filter(Q(published__gt=date_from) & Q(published__lt=date_to))
```
sirex:
```suggestion
        if date_from and date_to:
            datasets = datasets.filter(published__range=(date_from, date_to))
```

https://docs.djangoproject.com/en/3.2/ref/models/querysets/#range

Filtering by dates should be inclusive.

vitrina/datasets/views.py
```diff
@@ -14,3 +20,168 @@ class DatasetListView(ListView):
...
+        elif date_from:
+            datasets = datasets.filter(published__gt=date_from)
```
sirex:
```suggestion
            datasets = datasets.filter(published__gte=date_from)
```

vitrina/datasets/views.py
```diff
@@ -14,3 +20,168 @@ class DatasetListView(ListView):
...
+        elif date_to:
+            datasets = datasets.filter(published__lt=date_to)
```
sirex:
```suggestion
            datasets = datasets.filter(published__lte=date_to)
```

vitrina/datasets/views.py
```diff
@@ -14,3 +20,168 @@ class DatasetListView(ListView):
...
+    paginate_by = 20
+
+    def get_queryset(self):
+        query = self.request.GET.get('q')
+        date_from = self.request.GET.get('date_from')
+        date_to = self.request.GET.get('date_to')
+        status = self.request.GET.get('status')
+        tags = self.request.GET.getlist('tags')
+        categories = self.request.GET.getlist('category')
+        filter_dict = self.request.GET.dict()
```
sirex: Since this is going directly into query it will for sure generate a lot of 500 errors, when some bot passes some GET params. Let's be explicit and only include GET params, that we know are model parameters.

vitrina/datasets/views.py
```diff
@@ -14,3 +20,168 @@ class DatasetListView(ListView):
...
+    template_name = 'vitrina/datasets/list.html'
+    paginate_by = 20
+
+    def get_queryset(self):
+        query = self.request.GET.get('q')
+        date_from = self.request.GET.get('date_from')
+        date_to = self.request.GET.get('date_to')
+        status = self.request.GET.get('status')
+        tags = self.request.GET.getlist('tags')
+        categories = self.request.GET.getlist('category')
```
sirex: Could you move all filter parameters into a form? We should always validate user input, otherwise we will definitely run into problems.

vitrina/datasets/views.py
```diff
@@ -14,3 +20,168 @@ class DatasetListView(ListView):
...
+        selected_value = [] if multiple else None
+        value = self.request.GET.getlist(title) if multiple else self.request.GET.get(title)
+        if value:
+            selected_value = value
+            if is_int:
+                try:
+                    selected_value = [int(val) for val in value] if multiple else int(value)
+                except ValueError:
+                    pass
+        return selected_value
```
sirex: Let's keep View classes clean and do not add any helper methods. All view helper functions should go to `helpers.py`.

https://en.wikipedia.org/wiki/Separation_of_concerns
https://en.wikipedia.org/wiki/Single-responsibility_principle

vitrina/datasets/views.py
```diff
@@ -14,3 +20,168 @@ class DatasetListView(ListView):
...
+                for ch in children:
+                    count += filtered_queryset.filter(category=ch).count()
+                category_counts[category.pk] = count
+
+        extra_context = {
+            'num_found': filtered_queryset.count(),
+            'status_filters': [{
+                'key': key,
+                'title': value,
+                'query': "?%s%sstatus=%s" % (query, "&" if query else "", key),
```
sirex: It looks, that new filters are always appended on top of existing ones. I think, better way to handle this, is to have `query` as a dict and update it, replacing or appending previously set values. Then you can simply [urlencode](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) dict Also, please move this url building functionality into a separate function in `helpers.py` and add tests for it. The end result could look like this:

```suggestion
                'query': get_filter_url(query, 'status', key, append=False),
```

sirex (changes requested):

vitrina/datasets/views.py
```diff
@@ -1,16 +1,134 @@
...
 class DatasetListView(ListView):
     model = Dataset
-    queryset = Dataset.public.order_by('-published')
     template_name = 'vitrina/datasets/list.html'
     paginate_by = 20
 
+    def get_queryset(self):
+        datasets = Dataset.public.order_by('-published')
+        filter_form = DatasetFilterForm(self.request.GET)
+        filter_form.is_valid()
```
sirex, line 37: If form is not valid, an error page should be shown, instead of continuing even with invalid user input.

vitrina/helpers.py
```diff
@@ -0,0 +1,23 @@
...
+        selected_value = value
+        if is_int:
+            try:
+                selected_value = [int(val) for val in value] if multiple else int(value)
+            except ValueError:
+                return [] if multiple else None
+    return selected_value
+
+
+def get_filter_url(request, key, value, append=False):
```
sirex:
```suggestion
def get_filter_url(request: Request, key: str, value: Any, append: bool = False) -> str:
```

vitrina/helpers.py
```diff
@@ -0,0 +1,23 @@
+from urllib.parse import urlencode
+
+
+def get_selected_value(request, title, multiple=False, is_int=True):
```
sirex:
```suggestion
def get_selected_value(request: Request, title: str, multiple: bool = False, is_int: bool = True) -> Optional[List[Any]]:
```

sirex (changes requested):

sirex (changes requested):
I still see translations files. What I meant in previous comment, that translation files should net be in this pull request, now they are deleted from `devel`.

https://git-scm.com/docs/git-restore

Same thing applies to static files, static files should not by added to version control, they should be managed by an asset manager.

https://webpack.js.org/guides/getting-started/

adp-atea:
As I understand the last change will be done with another task, so putting to review.

vitrina/datasets/views.py
```diff
@@ -1,16 +1,145 @@
...
+            if category:
+                related_categories = get_related_categories(category, only_children=True)
+                datasets = datasets.filter(category__pk__in=related_categories)
+            if organization:
+                datasets = datasets.filter(organization=organization)
+            if frequency:
+                datasets = datasets.filter(frequency=frequency)
+            return datasets
+        else:
+            return HttpResponseServerError()
```
sirex: Please, respect the interface, `get_queryset` must return `QuerySet`, not `HttpResponse`. In order to handle this correctly, you need to move form validation into `get` method where, you can return `HttpResponse`. When you follow the interface, you don't have to resort to hacks like `hasattr(self.object_list, 'status_code')`.

Also, we need to return a [HttpResponseBadRequest](https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpResponseBadRequest), not a server error.

See also:

- https://github.com/django/django/blob/stable/3.2.x/django/views/generic/list.py#L141-L158
- https://docs.djangoproject.com/en/3.2/ref/class-based-views/generic-display/#listview
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

sirex (changes requested):

sirex (approved):

PR #132, opened 2022-08-25
94 peradresavimai

### LINKED ISSUE #94: Peradresavimai
Įgyvendintas patikrimas, kai tam tikras duotas URL nerandamas routingo lentelėje, ar toks URL nėra įrašytas į peradresavimų lentelę. Peradresavimų lentelėje nurodyta laikas, kada peradresavimas buvo sukurtas; URL pavadinimas (iš `urls.py`) iš kurio peradresuojama; URL kelias,`object_id`, `object_type` - generic ryšys su objektu į kurį peradresuota (žiūrėti [Generic relations](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#generic-relations). 
Modulis: `compat`.

---

Tais atvejais, kai tam tikras duotas URL nerandamas routingo lentelėje, reikia patikrinti ar toks URL nėra įrašytas į peradresavimų lentelė.

Peradresavimų lentelėje turi būti tokie laukai:

- `created` - laikas, kada peradresavimas buvo sukurtas.
- `name` - URL pavadinimas (iš `urls.py`) iš kurio peradresuojama, pavyzdžiui `dataset-details-old`.
- `path` - URL kelias, pavyzdžiui `/dataset/my-dataset`.
- `object_id`, `object_type` - generic ryšys su objektu į kurį reikia peradresuoti (žiūrėti [Generic relations](https://docs.djangoproject.com/en/3.2/ref/contrib/contenttypes/#generic-relations).

Visi peradresavimai turi būti `308 Permanent Redirect`.

- [ ] Jei randamas peradresavimo įrašas, tada tiesiog imamas objektas, pagal `object_id` ir `object_type` ir nukreipimas atliekamas į to objekto [get_absolute_url](https://docs.djangoproject.com/en/3.2/ref/models/instances/#get-absolute-url).

Modulis: `compat`.

adp-atea:


.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
```
sirex: Could you remove `.idea/` from version control? Usually everyone uses IDE with different settings, so I thing, pushing someones personal settings is not a good idea.

vitrina/redirections/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+
+
+class Redirections(models.Model):
+    objects = models.Manager()
```
sirex:
```suggestion
```

`objects` manager is added by default.

vitrina/redirections/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class StructureConfig(AppConfig):
+    name = 'vitrina.redirections'
+    label = 'vitrina_redirections'
+    verbose_name = _("Redirections")
```
sirex: This whole app should go into `compat` module, not `redirections`, we will need to add more backwards compatibility features, not just redirects.

vitrina/redirections/views.py
```diff
@@ -0,0 +1,11 @@
...
+from django.shortcuts import redirect
+
+
+def redirection_handler(request, exception):
+    try:
+        existing_url = Redirections.objects.filter(path=request.path)
+        new_url = existing_url.content_object.get_absolute_url
+        return redirect(new_url)
+    except:
+        return redirect('home')
```
sirex: If we can't handle redirect, then we should call original Django 404 handler.

vitrina/redirections/views.py
```diff
@@ -0,0 +1,11 @@
+from vitrina.redirections.models import Redirections
+from django.shortcuts import redirect
+
+
+def redirection_handler(request, exception):
+    try:
+        existing_url = Redirections.objects.filter(path=request.path)
+        new_url = existing_url.content_object.get_absolute_url
```
sirex: Please add tests, I don't thing this will work, because `redirect` does not accept callable objects.

https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#redirect

```suggestion
        new_url = existing_url.content_object
```

vitrina/redirections/models.py
```diff
@@ -0,0 +1,13 @@
...
+    created = models.DateTimeField(blank=True, null=True, auto_now_add=True)
+    name = models.TextField(blank=True, null=True)
```
sirex:
```suggestion
    name = models.CharField(max_length=255, blank=True)
```

We don't want to add extra burden of handling null values. Empty string is sufficient in our case.

vitrina/redirections/models.py
```diff
@@ -0,0 +1,13 @@
...
+    path = models.TextField(blank=True, null=True)
```
sirex:
```suggestion
    path = models.CharField(max_length=255, index=True)
```

This should be a required field. And we want to add index in order to speed up queries.

vitrina/urls.py
```diff
@@ -39,7 +40,10 @@
     path('', include('vitrina.classifiers.urls')),
     path('', include('vitrina.cms.urls')),
     path('', include('vitrina.translate.urls')),
+    path('', include('vitrina.redirections.urls')),
```
sirex: There are no urls, so there is not point including it.

vitrina/redirections/urls.py
```diff
@@ -0,0 +1,4 @@
+from django.urls import path
+
+urlpatterns = [
+]
```
sirex: There is no point keeping this file, since it is empty. I suggest removing it.

sirex (changes requested):
There are no tests.

tests/compat/test_redirections.py
```diff
@@ -0,0 +1,30 @@
...
+    c = Client()
+    response = c.get('/neegzistuoja/')
+    assert response.status_code == 404
+
+
+@pytest.mark.django_db
+def test_redirection_exists(app: DjangoTestApp):
+    custom_dataset = DatasetFactory()
+    ctype = ContentType.objects.get(model='dataset')
+    Redirections.objects.create(created=datetime.now(),
```
sirex: Do not ignore [warnings](https://github.com/atviriduomenys/katalogas/runs/8087420966?check_suite_focus=true#step:7:43), currently there are many warnings and using naive dates is one of the warnings.

```suggestion
    Redirections.objects.create(created=datetime.now(timezone.utc),
```

https://docs.python.org/3/library/datetime.html#aware-and-naive-objects

tests/compat/test_redirections.py
```diff
@@ -0,0 +1,30 @@
...
+                                name="labas",
+                                path="/labas/",
+                                content_type=ctype,
+                                object_id=custom_dataset.pk)
+    resp = app.get('/labas/')
+    assert resp.status == '308 Permanent Redirect'
```
sirex: Could you also check `resp.headers['location']`?

vitrina/compat/models.py
```diff
@@ -0,0 +1,12 @@
+from django.db import models
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.contrib.contenttypes.models import ContentType
+
+
+class Redirections(models.Model):
+    created = models.DateTimeField(blank=True, null=True, auto_now_add=True)
```
sirex:
```suggestion
    created = models.DateTimeField(blank=True, auto_now_add=True)
```

Is there are any reasons for `null=True`?

vitrina/compat/models.py
```diff
@@ -0,0 +1,12 @@
...
+    name = models.CharField(max_length=255, blank=True, null=True)
```
sirex:
```suggestion
    name = models.CharField(max_length=255, blank=True)
```

Is there are any reasons for `null=True`, why `blank=True` is not sufficient? There is a general rule, to never add nullable strings, because this will complicate handling strings. Nullable strings can be added if there is a good reason for it and empty strings have a meaning.

vitrina/compat/models.py
```diff
@@ -0,0 +1,12 @@
...
+    path = models.CharField(max_length=255, db_index=True)
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.PositiveIntegerField()
```
sirex:
```suggestion
    object_id = models.IntegerField()
```

Use same type as it is used for primary keys.

https://docs.djangoproject.com/en/3.2/ref/models/fields/#autofield

vitrina/compat/views.py
```diff
@@ -0,0 +1,13 @@
...
+
+def redirection_handler(request, exception):
+    try:
+        existing_url = Redirections.objects.get(path=str(request.path))
+        new_url = existing_url.content_object
+        response = HttpResponse(status=308)
+        response['location'] = new_url
+        return response
+    except:
+        return HttpResponseNotFound()
```
sirex:
```suggestion
        return page_not_found(request, exception)
```

Please, leave default behavior untouched.

https://docs.djangoproject.com/en/3.2/ref/views/#django.views.defaults.page_not_found

sirex (changes requested):
I'm really sorry, but I just found, that Django has exactly the same thing:

https://docs.djangoproject.com/en/3.2/ref/contrib/redirects/

So unfortunately, I think it will be best, to drop current implementation and just use `django.contrib.redirects`? Sorry, somehow, I missed this app before.

Also you can ignore all my comments for this pull request.

vitrina/settings.py
```diff
@@ -28,7 +28,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-ALLOWED_HOSTS = []
+ALLOWED_HOSTS = ['*']
```
sirex:
```suggestion
ALLOWED_HOSTS = ['data.gov.lt', 'staging.data.gov.lt']
```

Please provide explicit list of allowed hosts to prevent host header attacks.

https://docs.djangoproject.com/en/3.2/ref/settings/#allowed-hosts

Also you can add `localhost` if that was the reason for adding `'*'`.

sirex (changes requested):

sirex (approved):

PR #133, opened 2022-08-30
85 naudotojo prisijungimai ir registracija

### LINKED ISSUE #85: Naudotjo prisijungimai ir registracija
Naudojanto Django auth modulį.

adp-atea:
…gout, password reset and policy views


vitrina/users/forms.py
```diff
@@ -0,0 +1,156 @@
...
+        cleaned_data = super().clean()
+        first_name = cleaned_data.get('first_name', "")
+        last_name = cleaned_data.get('last_name', "")
+        password = cleaned_data.get('password', "")
+        email = cleaned_data.get('email', "")
+
+        if (len(first_name) < 3 or not first_name.isalpha()) or (len(last_name) < 3 or not last_name.isalpha()):
+            raise ValidationError(_("Vardas ir pavardė negali būti trumpesni nei 3 simboliai, negali turėti skaičių"))
+        if email and User.objects.filter(email=email).exists():
+            raise ValidationError(_("Naudotojas su tokiu el. pašto adresu jau egzistuoja"))
```
sirex: This is validated by `super().celan()`, but you should call it.

https://docs.djangoproject.com/en/3.2/topics/forms/modelforms/#overriding-the-clean-method

vitrina/users/views.py
```diff
@@ -0,0 +1,55 @@
...
+    def post(self, request, *args, **kwargs):
+        form = self.form_class(request.POST)
+        if form.is_valid():
+            user = form.save()
+            login(request, user, backend='django.contrib.auth.backends.ModelBackend')
+            return redirect('home')
+        # this is for email unique error that comes from model, so that there wouldn't be two of them,
+        # because there's already one in clean method
+        if "email" in form.errors:
+            form.errors.pop('email')
```
sirex: Email field has `unique=True`, so you don't need to validate that. Maybe removing email uniqueness validation from `RegisterForm` would solve this error duplication?

tests/users/test_views.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@pytest.fixture
+def user():
+    user = User.objects.create_user(email="test@test.com", password="test123")
+    return user
+
+
+@pytest.mark.django_db
+def test_login_with_wrong_credentials(client: Client, user):
```
sirex:
```suggestion
def test_login_with_wrong_credentials(client: Client, user: User):
```

User type annotations.

tests/users/test_views.py
```diff
@@ -0,0 +1,136 @@
...
+    email_error = "Naudotojas su tokiu el. pašto adresu jau egzistuoja"
+    password_error = "Slaptažodis turi būti ne trupesnis nei 8 simboliai, jį turi sudaryti raidės " \
+                     "ir skaičiai bei specialus simbolis"
+    terms_error = "Turite sutikti su naudojimo sąlygomis"
+    return {
+        'name_error': name_error,
+        'email_error': email_error,
+        'password_error': password_error,
+        'terms_error': terms_error
+    }
```
sirex: I don't see a point having this as a fixture? You can simply define it as a module level global variable.

tests/users/test_views.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@pytest.mark.django_db
+def test_register_with_short_name(client: Client, user, error_messages):
+    resp = client.post(reverse('register'), data={'first_name': "T", 'last_name': 'User', 'email': "test_@test.com",
+                                                  'password': "test123?", 'agree_to_terms': True})
+    assert list(resp.context['form'].errors.values()) == [[error_messages['name_error']]]
+    assert User.objects.count() == 1
```
sirex:
```suggestion
    assert User.objects.filter(email='test_@test.com').count() == 0
```

tests/users/test_views.py
```diff
@@ -0,0 +1,136 @@
...
+    return {
+        'name_error': name_error,
+        'email_error': email_error,
+        'password_error': password_error,
+        'terms_error': terms_error
+    }
+
+
+@pytest.mark.django_db
+def test_register_with_short_name(client: Client, user, error_messages):
```
sirex:
```suggestion
def test_register_with_short_name(client: Client, error_messages):
```

`user` fixture is not used in this test.

sirex: Could you also fix this for other tests?

tests/users/test_views.py
sirex: Could you also fix this for other tests?

vitrina/users/forms.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def get_user(self):
+        return self.user_cache
+
+
+class RegisterForm(ModelForm):
+    first_name = CharField(label=_("Vardas"), required=True, )
+    last_name = CharField(label=_("Pavardė"), required=True)
+    email = EmailField(label=_("El. paštas"), required=True, error_messages={})
+    password = CharField(widget=PasswordInput, label=_("Slaptažodis"), required=True)
```
sirex: Shouldn't there be a `password2` field, for checking if password was correctly entered two times?

https://docs.djangoproject.com/en/3.2/topics/auth/default/#django.contrib.auth.forms.UserCreationForm

vitrina/users/forms.py
```diff
@@ -0,0 +1,156 @@
...
+        email = cleaned_data.get('email', "")
+
+        if (len(first_name) < 3 or not first_name.isalpha()) or (len(last_name) < 3 or not last_name.isalpha()):
+            raise ValidationError(_("Vardas ir pavardė negali būti trumpesni nei 3 simboliai, negali turėti skaičių"))
+        if email and User.objects.filter(email=email).exists():
+            raise ValidationError(_("Naudotojas su tokiu el. pašto adresu jau egzistuoja"))
+        if len(password) < 8 or not any(i.isdigit() for i in password) or not any(i.isalpha() for i in password) \
+                or not any(not c.isalnum() for c in password):
+            raise ValidationError(_("Slaptažodis turi būti ne trupesnis nei 8 simboliai, "
+                                    "jį turi sudaryti raidės ir skaičiai bei specialus simbolis"))
```
sirex: Use standard password validation.

https://docs.djangoproject.com/en/3.2/topics/auth/passwords/#integrating-validation

vitrina/users/forms.py
```diff
@@ -0,0 +1,156 @@
...
+        user = super().save(commit=False)
+        user.set_password(self.cleaned_data["password"])
+        if user.version is None:
+            user.version = 1
+        if user.needs_password_change is None:
+            user.needs_password_change = False
+        if user.suspended is None:
+            user.suspended = False
+        if user.disabled is None:
+            user.disabled = False
```
sirex: Maybe all this could be set as default values on model?

tests/users/test_views.py
```diff
@@ -0,0 +1,136 @@
...
+    assert list(resp.context['form'].errors.values()) == [[error_messages['password_error']]]
+    assert User.objects.count() == 1
+
+
+@pytest.mark.django_db
+def test_register_with_password_with_no_spec_symbols(client: Client, user, error_messages):
+    resp = client.post(reverse('register'), data={'first_name': "Test", 'last_name': 'User', 'email': "test_@test.com",
+                                                  'password': "test1234", 'agree_to_terms': True})
+    assert list(resp.context['form'].errors.values()) == [[error_messages['password_error']]]
+    assert User.objects.count() == 1
```
sirex: There is no point using expensive high level tests, just for testing password validation. Test password validation function directly. Or better, use default Django password validation system.

https://docs.djangoproject.com/en/3.2/topics/auth/passwords/#module-django.contrib.auth.password_validation

sirex (changes requested):

vitrina/users/forms.py
```diff
@@ -0,0 +1,142 @@
...
+        cleaned_data = super().clean()
+        new_password = cleaned_data.get('new_password', "")
+
+        if self.user.check_password(new_password):
+            raise ValidationError(_("Slaptažodis negali būti toks pat"))
+        if len(new_password) < 8 or not any(i.isdigit() for i in new_password) \
+                or not any(i.isalpha() for i in new_password) \
+                or not any(not c.isalnum() for c in new_password):
+            raise ValidationError(_("Slaptažodis turi būti ne trupesnis nei 8 simboliai, "
+                                    "jį turi sudaryti raidės ir skaičiai bei specialus simbolis"))
```
sirex: Use standard password validation.

https://docs.djangoproject.com/en/3.2/topics/auth/passwords/#integrating-validation

tests/users/test_views.py
```diff
@@ -0,0 +1,110 @@
...
+    user = User.objects.create_user(email="test@test.com", password="test123")
+    return user
+
+
+@pytest.mark.django_db
+def test_login_with_wrong_credentials(csrf_exempt_django_app: DjangoTestApp, user: User):
+    resp = csrf_exempt_django_app.post(reverse('login'), {
+        'email': "test@test.com",
+        'password': "wrongpassword"
+    })
```
sirex: It is better to use a webtest form posting, this way, you won't need csrf excempt and more things will be tested, for example actual form rendering in templates will be tested, making sure if all the fields are actually in the form.

Could you also switch to webtest form handling in all other tests in this file?

https://docs.pylonsproject.org/projects/webtest/en/latest/forms.html

sirex (changes requested):

sirex (approved):

PR #140, opened 2022-09-02
136 move to webtest

### LINKED ISSUE #136: Naudoti webtest testuose
Siekiant užtikrinti geresnį kodo padengimą testais, naudojamas webtest karkasas, kurio pagalba testuojama ne tik atskiros užklausos, bet ir naudotojo sąsajos elementai.

adp-atea:


sirex (changes requested):
I'm really sorry, but this task is about moving tests, that user Django client into WebTest client. But what is done here is not exactly what was in the task.

There are two different ways to write tests:

- old unittest style tests, with classes
- new pytest style tests, with functions

All our tests are written using the new pytest style tests and that is good, we don't want to change that. We use simple functions instead of class methods and we use pytest fixtures instead of TestCase class setUp and tearDown methods.

What I see here, that all tests are refactored into old unittest style, but this is not, what this task is about. This task is about changing some tests, that use `client` instead of `app` fixture. Here is an example from `tests/test_view_count.py`:

```python
@pytest.mark.django_db
def test_view_count_project(app: DjangoTestApp):
    # [...]
    client = Client()
    # [...]
    client.force_login(user1)
    # [...]
```

Here, `app` fixture is loaded, but is not used for some reason, because a `client` instance is created and the testing continue with `client`, instead of `app`.

So what we need to do here, is to change `client` into `app` in all tests, that does thing like in this example.

Sorry for the confusion.

sirex (approved):

PR #143, opened 2022-09-09
83 auth beckendas adp slaptažodžiams

### LINKED ISSUE #83: Auth beckendas adp slaptažodžiams
Sukurti naują Django auth backend, kuris atpažintu, dabartinius slaptažodžių hashus.

adp-atea:
Django allows bcrypt password hashing, so I think writing another auth backend is unnecessary. The only thing is that Django adds hash algorithm name to the start of password field, so I added migration, that would fix existing passwords.


vitrina/users/migrations/0004_auto_20220909_1301.py
```diff
@@ -0,0 +1,23 @@
...
+from django.db import migrations
+from django.db.models import Q
+
+
+def fix_passwords(apps, schema_editor):
+    User = apps.get_model('vitrina_users', 'User')
+    users = User.objects.exclude(Q(password__startswith='bcrypt$') | Q(password__startswith='pbkdf2_sha256$'))
+    for user in users:
+        user.password = "bcrypt$%s" % user.password
+        user.save(update_fields=['password'])
```
sirex, line 12: It would be nice to have a test for this.

https://github.com/wemake-services/django-test-migrations#pytest

sirex (changes requested):

vitrina/users/migrations/0004_auto_20220909_1301.py
adp-atea: Tried using django-test-migrations, however since we don't have migrations for all modules, running tests gives errors unless we add --no-migrations to test command. When we add this parameter django-test-migrations tests are just skipped. I found another way to test the migration, if that is suitable.

sirex (approved):

PR #156, opened 2022-09-15
Added filtering

### LINKED ISSUE #149: Pataisyti Dataset.public managerį
Modifikuotas `PublicDatasetManager` filtravimas atrenkant tuos duomenų rinkinius, kurie yra matomi viešai.

---

Šiuo metu `PublicDatasetManager` nedaro jokio filtravimo, nors turėtu atrinkti tik tuos duomenų rinkinius, kurie yra matomi viešai.

Pavyzdžiui dabar atrenkama 2006 rinkiniai, nors data.gov.lt rodo 1782.

adp-atea:
Made it so PublicDatasetManager would show all datasets that don't have is_public set to False


vitrina/datasets/managers.py
```diff
@@ -1,6 +1,7 @@
 from django.db import models
+from django.db.models import Q
 
 
 class PublicDatasetManager(models.Manager):
     def get_queryset(self):
-        return super().get_queryset()
+        return super().get_queryset().filter(~Q(is_public=False))
```
sirex:
```suggestion
        return super().get_queryset().filter(is_public=True)
```

I think there is no need to use `~Q` here.

Also, this filter is not complete, for example it does not filter out deleted datasets. Please make sure, that this filter does exactly the same thing as in current data.gov.lt version.

sirex (changes requested):

vitrina/datasets/managers.py
```diff
@@ -1,6 +1,7 @@
 from django.db import models
+from django.db.models import Q
 
 
 class PublicDatasetManager(models.Manager):
     def get_queryset(self):
-        return super().get_queryset()
+        return super().get_queryset().filter(is_public=True).filter(~Q(deleted=True))
```
sirex: This is not, how it works currently. Here is a code sample, how datasets are filtered now:

```java
        if (d.getDeleted() != null || d.getDeletedOn() != null) {
            return;
        }

        if (d.getPublic() == null || !d.getPublic()) {
            return;
        }

        // todo check this one
        if (d.getOrganizationId() != null) {
            Optional<Organization> byId = organizationRepository.findById(d.getOrganizationId());
            if (byId.isPresent()) {
                if (byId.get().getPublic() != null) {
                    if (!byId.get().getPublic()) {
                        return;
                    } else {
                        indexOrganization(byId.get());
                        organizationCore.commit();
                    }
                }
            }
        }
```

Also you must always handle NULL values explicitly, because if you don't they will be ignored and will show up in the results, so correct query would be `.filter(deleted__isnull=False)`.

And please, add tests for this, because this is second incorrect attempt.

sirex (changes requested):

vitrina/datasets/managers.py
```diff
@@ -3,4 +3,8 @@
 
 class PublicDatasetManager(models.Manager):
     def get_queryset(self):
-        return super().get_queryset()
+        all_datasets = super().get_queryset()
+        public_datasets = all_datasets.filter(is_public__isnull=False).filter(is_public=True)
```
sirex:
```suggestion
        public_datasets = all_datasets.filter(is_public=True)
```

`is_public` is not nullable, so there is no point checking for NULL values.

sirex (changes requested):
Migrations do not work.

```
$ poetry run python manage.py migrate
  Applying vitrina_datasets.0001_initial...
Traceback (most recent call last):
  File "django/db/backends/utils.py", line 82, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.DuplicateTable: relation "dataset" already exists
```

sirex (approved):

PR #170, opened 2022-09-27
151 duomenų rinkinių filtras naudojant django-haystack

### LINKED ISSUE #151: Duomenų rinkinių filtras naudojant django-haystack
Modifikuotas duomenų rinkinių filtras naudojant [django-haystack](http://haystacksearch.org/) ir atnaujintas README dėl paieškos indekso atnaujinimo.

---

Perdaryti duomenų rinkinių filtrą naudojant [django-haystack](http://haystacksearch.org/) su pasirinktu (Solr ar ElasticSearch) backendu.

- [ ] Filtrus su įrašu skaičiais įgyvendinti naudojanti [Faceting](https://django-haystack.readthedocs.io/en/master/faceting.html) rodyti top dešimt rūšiuojant pagal count.
- [ ] Filtrai pagal žymes neturi rodyti tuščios žymės kaip filtro
- [ ] Pereinant į sekantį puslapį neturi nusimušti filtras
- [ ] Paieškai pagal raktinius žodžius taip pat naudoti Haystack
- [ ] Facetai turėtu būti surūšiuoti pagal count skaičių.
- [ ] Įtraukti Solr ar ElasticSearch į docker-compose.
- [ ] Atnaujinti README nurodant kaip atnaujinti paieškos indeksą.

adp-atea:


.github/workflows/run_tests.yml
```diff
@@ -21,11 +21,26 @@ jobs:
       - name: Install Poetry
         uses: snok/install-poetry@v1
 
+      - name: Configure sysctl limits
+        run: |
+          sudo swapoff -a
+          sudo sysctl -w vm.swappiness=1
+          sudo sysctl -w fs.file-max=262144
+          sudo sysctl -w vm.max_map_count=262144
```
sirex, line 30: If you found this somewhere on the internet, could you add a link to it?

.github/workflows/run_tests.yml
```diff
@@ -21,11 +21,26 @@ jobs:
...
+          stack-version: 7.8.1
+
       - name: Install dependencies
         run: poetry install
 
       - name: Create DB image
         run: docker-compose -f docker-compose-test.yml up -d
 
+      - name: Check ports
+        run: netstat -tulpn
```
sirex: Do we need this?

vitrina/settings.py
```diff
@@ -255,3 +258,22 @@
...
+        },
+    }
+else:
+    HAYSTACK_CONNECTIONS = {
+        'default': {
+            'ENGINE': 'haystack.backends.elasticsearch7_backend.Elasticsearch7SearchEngine',
+            'URL': 'http://127.0.0.1:9200/',
+            'INDEX_NAME': 'haystack'
+        },
+    }
```
sirex: Use [search_url](https://django-environ.readthedocs.io/en/latest/types.html#environ-env-search-url).

vitrina/settings.py
```diff
@@ -38,6 +39,7 @@
 
 ALLOWED_HOSTS = ['localhost', '127.0.0.1', 'data.gov.lt', 'staging.data.gov.lt']
 
+TEST = any(['test' in arg for arg in sys.argv])
```
sirex: Settings is not the place to recognize if we are running tests. Settings should be modified trough environment variables, using [django-environ](https://django-environ.readthedocs.io/).

```suggestion
```

tests/datasets/test_views.py
```diff
@@ -82,91 +86,104 @@ def test_distribution_preview(app: DjangoTestApp, dataset_detail_data):
...
 
+def rebuild_index():
+    ui = UnifiedIndex()
+    index = DatasetIndex()
+    ui.build(indexes=[index])
+    haystack.connections["default"]._index = ui
+
+    backend = haystack.connections["default"].get_backend()
+    backend.clear()
+    backend.update(index, Dataset.objects.all())
```
sirex: Use [pytest-django-haystack](https://pypi.org/project/pytest-django-haystack/).

tests/orgs/test_views.py
```diff
@@ -38,8 +42,16 @@ def test_organization_members_tab(app: DjangoTestApp, data_for_tabs):
 
 @pytest.mark.django_db
 def test_organization_dataset_tab(app: DjangoTestApp, data_for_tabs):
+    ui = UnifiedIndex()
+    index = DatasetIndex()
+    ui.build(indexes=[index])
+    haystack.connections["default"]._index = ui
+    backend = haystack.connections["default"].get_backend()
+    backend.clear()
+    backend.update(index, Dataset.objects.all())
```
sirex: Use [pytest-django-haystack](https://pypi.org/project/pytest-django-haystack/).

sirex (changes requested):

.github/workflows/run_tests.yml
adp-atea: https://github.com/elastic/elastic-github-actions/tree/master/elasticsearch

tests/datasets/test_views.py
adp-atea: Could not add pytest-django-haystack, because it uses method getfuncargvalue, which is not available anymore. It was changed to getfixturevalue.  So I copied the code from  pytest-django-haystack and added haystack marker to conftest.py with changed method name.

.github/workflows/run_tests.yml
sirex: Can you add it to the `run_tests.yml` file as a comment?

tests/conftest.py
```diff
@@ -21,3 +21,38 @@ def app(django_app):
...
+        from pytest_django.lazy_django import skip_if_no_django
+        from django.core.management import call_command
+        request.getfixturevalue('db')
+
+        # optional haystack connection parameter
+        # if specified, pass to clear_index and rebuild_index
+        connection = marker.kwargs.get('connection', None)
+        index_args = {}
+        if connection:
+            index_args['using'] = connection
```
sirex:
```suggestion
        index_args['using'] = 'test'
```

This way we will be able to get rid of `test_settings.py`, by simply adding another `'test'` connection alongside `'default'` as it is done with database configuration settings.

tests/test_helpers.py
```diff
@@ -4,59 +4,59 @@
...
+    def __init__(self, selected_facets):
+        self.selected_facets = selected_facets
+
+
 @pytest.mark.django_db
-def test_get_selected_value(rf: RequestFactory):
-    request = rf.get('/', {'key': 'value'})
-    selected_value = get_selected_value(request, 'key', is_int=False)
+def test_get_selected_value():
+    facet_form = MockFacetForm(['key_exact:value'])
```
sirex:
```suggestion
    facet_form = Mock(selected_facets=['key_exact:value'])
```

Simply use the [Mock](https://docs.python.org/3/library/unittest.mock.html#the-mock-class) class. Please do the same thing in other places and remove `MockFacetForm` class.

vitrina/helpers.py
```diff
@@ -2,26 +2,31 @@
...
+        try:
+            selected_value = [int(val) for val in selected_value] if multiple else int(selected_value)
+        except ValueError:
+            return [] if multiple else None
     return selected_value
 
 
-def get_filter_url(request: WSGIRequest, key: str, value: str, append: bool = False) -> str:
+def get_filter_url(request: WSGIRequest, key: str, value: str) -> str:
     query_dict = dict(request.GET.copy())
```
sirex, line 28: You need to clean `page`, because I get `Invalid page` errors, when I go to the last page, then click on a faced, which returns fewer pages, but since `page` is not cleared, I get an Invalid page error.

Could you also write a test for this case.

vitrina/datasets/views.py
```diff
@@ -3,180 +3,62 @@
...
-        return HttpResponseBadRequest(_("Klaida"))
-
     def get_queryset(self):
-        datasets = Dataset.public.order_by('-published')
-        if is_org_dataset_list(self.request):
-            organization = get_object_or_404(
-                Organization,
-                pk=self.kwargs['pk'],
-            )
-            datasets = datasets.filter(organization=organization)
```
sirex, line 57: By removing this, you broke `organization-datasets` view, could you add tests for this case and fix the issue?

Currently `organization-datasets` does not filter by organization, but shows all datasets, from all organizations.

sirex (changes requested):

tests/conftest.py
adp-atea: I have tried this, but the problem is that settings are always loaded with 'default' value. Then when the tests are run, it starts to use 'test', but what happens is that 'default' and 'test' connections mix up and the search index has data from both development database and test database, which causes tests to fail and then we also need to rebuild the index  for development after running the tests, because it has test data.

tests/orgs/test_views.py
```diff
@@ -44,10 +45,10 @@ def test_organization_members_tab(app: DjangoTestApp, data_for_tabs):
     assert list(resp.html.find("li", class_="is-active").a.stripped_strings) == ["Organizacijos nariai"]
 
 
-@pytest.mark.django_db
+@pytest.mark.haystack
 def test_organization_dataset_tab(app: DjangoTestApp, data_for_tabs):
```
sirex:
```suggestion
def test_organization_dataset_tab(app: DjangoTestApp, data_for_tabs):
```

Is suggest, to avoid reusing data fixtures, better user factories every time, because most of the time, you need different data. Like in this, case, in order to reproduce [the issue](https://github.com/atviriduomenys/katalogas/pull/170#discussion_r989791578), you need to add at least two datasets, each with different organization and then use the search, to check if only one is shown.

vitrina/datasets/views.py
sirex: This is still not fixed.

vitrina/helpers.py
sirex: This is still not fixed. Please add a tests, that reproduces this issue.

tests/conftest.py
sirex: I think, in order to make this work, you need to specify a different connection in settings, here:

```python
HAYSTACK_CONNECTIONS = {
    'default': env.search_url(),
    'test': env.search_url(default='elasticsearch://localhsot:9200/test'),
}
```

adp-atea: Yes, I tried it by adding 'test' connection to the settings, but result was as I said before.

sirex: I fixed it for you.

sirex (approved):

PR #162, opened 2022-09-21
117 turinio vertinimas spaudžiant „Patinka“

adp-atea:

PR #174, opened 2022-09-28
Form, views, tests, some fixes to other places

### LINKED ISSUE #154: Šaltinio forma
Modifikuotas duomenų rinkinio peržiūros langas, pridedant mygtuką „Naujas šaltinis“, kuris atidaro naujo šaltinio formą. Lango funkcionalumas konfigūruojamas dinamiškai, priklausomai nuo vartotojo rolės ir teisių.

--- 

- [ ] - [x] Duomenų rinkinio [details puslapyje](#68), duomenų šaltinių lentelės viršuje pridėti mygtukus „Naujas šaltinis“, kuris atidaro naujo šaltinio formą.
- [x] Nuorodos ir failo įkėlimo formas turi būti viena, jei užpildyta nuoroda, rodoma, kaip nuoroda, jei įkeltas failas, rodoma, kaip failas. Gali būti užpildytas tik failas arba tik nuoroda, bet ne abu.
- [x] Naujo šaltinio forma turėtu atitikti dabartinę šaltinio formą.
- [x] Teisės:  `user.is_staff`, `dataset.organization == user.organization` ir `dataset.manager = user`.
- [x] Šaltinių sąraše, prie kiekvieno šaltinio pridėti mygtuką „Keisti“, mygtuką rodyti jei naudotojas turi reikiamas teises. Mygtukas atidaro šaltinio keitimo formą.
- [x] Šaltinių sąraše, prie kiekvieno šaltinio pridėti mygtuką „Trinti“, jį paspaudus ištrinamas šaltinis. Mygtukas rodomas turint reikiamas teises. Rodom config dialogą (gali būti js alert).

adp-atea:


sirex (changes requested):
https://github.com/atviriduomenys/katalogas/pull/162#pullrequestreview-1123703725

sirex (changes requested):
I did several changes, to fix the distribution form, I think this way we can move faster. But my changes are not complete, so I would like to ask you to finish with refactoring of some places, adding schema and data migrations.

Where is a list of things that needs to be done:

- [x] `DistributionFormat.url_format` was renamed to `format`. Previously it was used only for URL type distributions, but now it is needed for both URL and FILE distributions. Please finish renaming this field. And also add data migration, by selecting a `Format` instance for `type=FILE`. To do that, you need to detect correct format from `mime_type` and `filename` extension. You can find and example, how I was doing it [here](https://nbviewer.org/gist/sirex/90b291c559bafed8295f61606ea4fa6a#Brandos-lygis).
- [x] `DistributionFormat.url` was renamed to `download_url`, to match DCAT naming and differentiate from `access_url`, which was introduced quite recently. Please finish the renaming.
- [x] `DistributionFormat.filename` was renamed to `file`, since this is no longer just a file name, but a `File` object.
- [x] `DistributionFormat` `.region` and `.municipality` types changed from `CharField` to `ForeignKey`, please finish refactoring other places where this was used and also add a schema and data migration. For data migration, you need to replace `region` and `municipality` values from titles to primary keys.
- [x] I think `DistributionFormat` fields `type` and `mime_type` are no longer needed, but do not delete them, just move down the field list and add a comment `Deprecated` above those fields.
- [x] Added [crispy_bulma](https://github.com/ckrybus/crispy-bulma) package in order to have a proper form styling, previous implementation was a mess. I did refactored `DatasetResourceForm` to accommodate this change, but could you please do the same thing to all other forms, where `FormHelper` is used?

I know, that is a lot of additional changes, but unfortunately, `DistributionFormat` was a painful places, and it needs to be fixed.

Since these are additional changes, I'm raising existing estimate from 12 to 24 hours.

vitrina/resources/models.py
```diff
@@ -40,70 +43,125 @@ class DatasetDistribution(models.Model):
...
+            'Saugykloje ar kitoje vietoje, pateikiant tiesioginę duomenų '
+            'atsisiuntimo nuorodą. Tačiau nedidelės apimties (iki 5Mb) '
+            'duomenų failus, galima talpinti ir kataloge.'
+        ),
+    )
+
+    geo_location = models.CharField(
+        max_length=255,
+        blank=True,
+        null=True,
```
sirex:
```suggestion
```

Char fields should never be null, unless there is a good reason for that. Please change this also for `period_start` and `period_end`.

Also, when adding a form with empty `geo_location`, `period_start` or `period_end` I get this error:

```python
Traceback (most recent call last):
  File "django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.NotNullViolation: null value in column "geo_location" of relation "dataset_distribution" violates not-null constraint
DETAIL:  Failing row contains (10938, 2022-10-10 18:15:26.796594+00, 2022-10-10 18:15:26.796629+00, 1, , null, , 
null, null, null, null, null, null, , URL, http://example.com/data.csv, 1254, null, null, null, 1, , null).
```

It looks, that model says null=True, but migrations things it's null=False. Anyway, CharField, should almost never be null=True. Empty string fully satisfies emptiness, there is no need for special kind of emptiness for strings.

vitrina/resources/migrations/0007_files_fix.py
```diff
@@ -0,0 +1,56 @@
...
+    DatasetDistribution = apps.get_model("vitrina_resources", "DatasetDistribution")
+    Format = apps.get_model("vitrina_resources", "Format")
+
+    for resource in DatasetDistribution.objects.all():
+        if resource.file:
+            mime_type = mimetypes.guess_type(str(resource.file))
+            detected_format = Format.objects.filter(mimetype=mime_type).first()
+            resource.format = detected_format
+        if resource.type == 'URL':
+            url_format = Format.objects.filter(extension='URL').first()
```
sirex: For URLs you actually need to do a real request and get format from response content type, like in example here:

```python
    status_code = 0
    content_type = ''
    try:
        resp = requests.get(url, stream=True, timeout=1)
    except requests.RequestException:
        pass
    else:
        if resp.status_code == 200:
            content_type = resp.headers.get('content-type')
            if ';' in content_type:
                content_type = content_type.split(';', 1)[0]
```

The you need to map received content type with a format:

```python
{
    'application/rdf+xml': 'RDF',
    'text/csv': 'CSV',
    'application/json': 'JSON',
    'application/geo+json': 'JSON',
    'application/xml': 'XML',
    'text/xml': 'XML',
    'application/vnd.ms-excel': 'XLSX',
    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet': 'XLSX',
    'text/html': 'HTML',
    'text/plain': 'URL',
    'application/zip': 'URL',
    '': 'URL',
}
```

And only fallback to URL if there is no way, to detect which format it actually is.

sirex (changes requested):

vitrina/resources/migrations/0007_files_fix.py
sirex: By the way, I tink it is better to move this migration into a script, because this migration is most likely will take a long time to complete.

sirex (approved):

PR #178, opened 2022-09-29
176 nuasmenintas dumpas

### LINKED ISSUE #176: Nuasmenintas dumpas
Pritaikyta Django management commanda, kuri nuasmenina duomenis, slaptažodžiai pakeist į "secret".

---

Django management commanda, kuris nuasmenina duomenis.

- Obfuksuoti userių vardus, telefonus, emailsu ir pan. Galima naudoti Faker, reikšmėms pakeisti.
- Pakeisti slaptažodžius į `secret`.

adp-atea:


sirex (changes requested):
I added some changes, but the issue is, that anonymization must be done on initial data, before Django migrations are applied, because we want to know if migrations will work.

I suggest moving this out of Django management command into `scripts/anonymize.py` and do the anonymization using [dataset](https://dataset.readthedocs.io/).

The process should look like this:

1. Move data from MySQL to PostgreSQL using pgloader.
2. Run `poetry run python scripts/anonymize.py` on the PostgreSQL database.
3. Run `poetry run manage.py migrate`.

This way we will have more reliable data for migrations.

adp-atea:
Note that because we're anonymizing data before migrations,  we should change user password length to be longer in MySQL dump file by hand, otherwise anonymize script will not be able to save changed passwords.

scripts/anonymize.py
```diff
@@ -0,0 +1,57 @@
...
+        users = {}
+        for user in db['user'].all():
+            fake = users[user['email']] = UserFactory.build()
+            data = {
+                'id': user['id'],
+                'first_name': fake.first_name,
+                'last_name': fake.last_name,
+                'email': fake.email,
+                'phone': fake.phone,
+                'password': make_password("secret")
```
sirex: We should use exactly the same hashing algorithm as it was used before the migration. We want to be as close as possible to the original production environment.

resources/adp-dev.sql
```diff
@@ -1207,7 +1207,7 @@ CREATE TABLE `user` (
   `first_name` varchar(255) COLLATE utf8mb4_lithuanian_ci DEFAULT NULL,
   `last_login` datetime(6) DEFAULT NULL,
   `last_name` varchar(255) COLLATE utf8mb4_lithuanian_ci DEFAULT NULL,
-  `password` char(60) COLLATE utf8mb4_lithuanian_ci DEFAULT NULL,
+  `password` char(255) COLLATE utf8mb4_lithuanian_ci DEFAULT NULL,
```
sirex: We should stick to the original password hashing algorithm and should not change the schema.

sirex (changes requested):

sirex (changes requested):
I see that depending on factories is not a good idea, because `Representative` and its factory will change significantly, so we can't relay on factories.

Could you remove dependency from factories and generate names, emails, etc. using [Faker](https://faker.readthedocs.io/en/master/providers/faker.providers.person.html) directly?

sirex (approved):

PR #183, opened 2022-09-30
Removed --no-migrations

### LINKED ISSUE #167: Ištrinti --no-migrations testuose
Modifikuoti CI metu leidžiami testai, artimi gamybinei aplinkai, pašalinant  `--no-migrations` argumentą bei įkėlus PostgreSQL schemos dumpą, sukurtą iš MySQL dumpo.

---

Šiuo metu CI metu leidžiami testai, nevykdo migracijų, o klaidų migracijose pasitaiko labai daug. Šiuo metu testų eilutė atrodo taip:

    poetry run pytest -vvra --no-migrations --tb=short --cov=vitrina --cov=report=term-missing tests

Reikėtų pašalinti `--no-migrations` argumentą.

Kitas trūkumas, kad testai leidžiami ant tuščios duomenų bazės, nors turėtu būti naudojama duomenų bazės schema perkelta iš MySQL. `resources` kataloge reikėtų įkelti PostgreSQL schemos dumpą, sukurtą iš MySQL dumpo, naudojant instrukcijas pateiktas `README.rst` faile. Tada PostgreSQL duomenų bazė testuose turėtu užkrauti pradinį dumpą, o testai turėtu būti vykdomi su `--reuse-db` argumentu, kad kiekvienas testas būtų vykdomas nuo pradinės būsenos, o ne nuo tuščios duomenų bazės.

Tai gali labai sulėtinti testus, todėl, kitas variantas, tiesiog padaryti atskirą testą, kuriame būtų paleidžiamos migracijos ant duomenų bazės su schema perkelta iš MySQL.

Esmė tokia, kad testai turi veikti sąlygomis, kiek įmanoma artimesnėmis gamybinei aplinkai.

adp-atea:
To achieve this django_db_setup needed to be overwritten to generate base database from adp-dev dump file and to call migrations.


tests/conftest.py
```diff
@@ -21,3 +27,39 @@ def app(django_app):
...
+        with django_db_blocker.unblock():
+            with connection.cursor() as cursor:
+                cursor.execute(sqlFile)
+            call_command('migrate')
+        file.close()
+
+    yield
+
+    for conn in connections.all():
+        conn.close()
```
sirex: I don't understand, why we need this? [pytest-django](https://pytest-django.readthedocs.io/en/latest/database.html) does exactly that.

.github/workflows/run_tests.yml
```diff
@@ -28,4 +28,4 @@ jobs:
         run: docker-compose -f docker-compose-test.yml up -d
 
       - name: Test with pytest
-        run: poetry run pytest -vvra --no-migrations --tb=short --cov=vitrina --cov=report=term-missing tests
\ No newline at end of file
+        run: poetry run pytest -vvra --reuse-db --tb=short --cov=vitrina --cov=report=term-missing tests
```
sirex: I guess, there is no point in `--reuse-db` in CI, because in CI whole server is recreated from scratch and there is no database to reuse. This option is only makes sense in development environment.

pytest.ini
```diff
@@ -1,3 +1,4 @@
 [pytest]
+addopts = --reuse-db --migrations
```
sirex: I don't suggest enforcing these options for everyone. So it is better to delete this line and leave option to add these options case by case.

vitrina/settings.py
```diff
@@ -151,7 +152,15 @@
...
-    'default': env.db(),
+    'default': {
+        'ENGINE': 'django.db.backends.postgresql',
+        'NAME': 'adp-dev',
+        'TESTNAME': 'test_adp_dev',
+        'USER': 'adp',
+        'PASSWORD': 'secret',
+        'HOST': '127.0.0.1',
+        'PORT': '5432',
+    },
```
sirex: Please leave `env.db()`. we want to configure database via environment variables. And this was specifically added as a part of a task.

sirex (changes requested):
Please move `tests/resources/static/db/adp-dev.sql` to `resources/adp-pg.sql`.

tests/conftest.py
```diff
@@ -21,3 +27,34 @@ def app(django_app):
...
+    run_sql = partial(_run_sql, settings.DATABASES['default'])
+    run_sql(f'DROP DATABASE IF EXISTS {test_db_name}')
+    run_sql(f'CREATE DATABASE {test_db_name}')
+    with open(settings.BASE_DB_PATH, 'r') as file:
+        sqlFile = file.read()
+        with django_db_blocker.unblock():
+            with connection.cursor() as cursor:
+                cursor.execute(sqlFile)
+            call_command('migrate')
+        file.close()
```
sirex: Could you move this out of `conftest.py` into docker-compose.yml, similarly like it was done with MySQL?

See https://github.com/docker-library/docs/blob/master/postgres/README.md#initialization-scripts

sirex (changes requested):

tests/conftest.py
adp-atea: Would need some guidance on how to do that, as per recent commits I have tried it, but nothing seems to work if I have my original code commented out.

sirex: Maybe this would work?

```sh
# Load initial database schema
psql -h localhost -U adp -d adp-dev < resources/adp-pg.sql

# Run migrations
poetry run python manage.py migrate

# Rename database to match test database name
psql -h localhost -U adp postgres -c 'alter database "adp-dev" rename to "test_adp-dev";'

# And run tests with `--reuse-db` on original database with migrations applied on top of it
poetry run pytest -vvxra --tb=short --reuse-db tests
```

We need to only run these commands on CI, to reproduce exactly the same conditions as in production setting. And by doing that I already found an issue:

```
psycopg2.errors.UndefinedColumn: column usecase_dataset_ids.id does not exist
```

Here, `UsercaseDatasetIds` is incorrectly defined, maybe it is possible to fix it by defining a proper [ManytoManyField.through](https://docs.djangoproject.com/en/3.2/ref/models/fields/#django.db.models.ManyToManyField.through), but it looks, that this too requires `id` field, if this is the case, the probably only way to fix it is through migrations.

sirex (changes requested):

tests/conftest.py
adp-atea: Since we are running migrations manually, i added --no-migrations and --reuse-db to the tests.

sirex (approved):

PR #248, opened 2022-12-15
236 kategorijų filtras paga lygmenis

### LINKED ISSUE #236: Kategorijų filtras paga lygmenis
Duomenų rinkinių sąraše, kategorijų filtras turi rodyti tik pasirinktą kategoriją, visas pasirinktos kategorijos subkategorijas ir visas pasirinktos kategorijos tėvines kategorijas. Jei nėra parinkta kategorija, tada rodomos tik aukščiausio lygmens kategorijos.

Duomenų rinkinių skaičius, atitinkantis kategorijas, turi apimti ir visų po konkrečia kategorija esančių subkategorijoms priskirtų rinkinių bendrą skaičių.

Šiuo metu rodomos visos kategorijos.

adp-atea:


vitrina/datasets/models.py
```diff
@@ -153,6 +153,16 @@ def get_absolute_url(self):
     def get_tag_list(self):
         return list(self.tags.all().values_list('name', flat=True))
 
+    @property
```
sirex: Properties are supposed to be very light weight, but here is an expensive method, with multiple queries to database. Remove `@property` decorator and make it a method as it should be.

vitrina/datasets/models.py
```diff
@@ -153,6 +153,16 @@ def get_absolute_url(self):
...
+    def parent_category(self):
+        categories = []
+        if self.category:
+            category = self.category
+            while category.parent:
+                category = category.parent
+            categories.append(category.pk)
+        return categories
```
sirex: One of the reasons why we use treebeard, is to avoid expensive code like this one, involving multiple queries. Please use a treebeard method instead.

sirex (approved):

PR #284, opened 2023-03-24
281 move format data migration out of migration files

### LINKED ISSUE #281: Move format data migration out of migration files
Move format fixing migration out of migrations into a separate script, because this migration take very long time to complete.

https://github.com/atviriduomenys/katalogas/blob/9349239ae43534b87ef2a8a0f28ec97033c0079d/vitrina/resources/migrations/0007_files_fix.py#L193

It would be great if requests would be cached to a specified locations, so that running it repeatedly, requests, that were already cached, would read data from the cache. For caching you can use [requests-cache](https://pypi.org/project/requests-cache/) library.

See other scripts in `scripts/` directory.

Also, check other migrations if there are other slow migrations, maybe they should be moved to a script too?

Basically we need to fix migrations, to make then run fast.

Faustasm:


scripts/fix_files.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+class _Result(NamedTuple):
+    dist: models.Model  # DatasetDistribution
+    format: str
+
+
+def main():
+    Dist = apps.get_model("vitrina_resources", "DatasetDistribution")
+    Format = apps.get_model("vitrina_resources", "Format")
```
sirex: Why can't we just import modes?

scripts/fix_files.py
```diff
@@ -0,0 +1,157 @@
...
+
+    formats = {
+        f.extension: f
+        for f in Format.objects.all()
+    }
+
+    client = CachedSession('demo_cache')
```
sirex: Cache location should be configurable via command line arguments.

scripts/fix_files.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+    dists_amount = len(Dist.objects.all())
```
sirex:
```suggestion
    dists_amount = Dist.objects.count()
```

sirex (changes requested):

scripts/fix_files.py
```diff
@@ -0,0 +1,153 @@
...
+    format: str
+
+
+def main(cache_location: str = Argument(..., help="File name to cache requessts")):
+    formats = {
+        f.extension: f
+        for f in Format.objects.all()
+    }
+
+    client = CachedSession(cache_location)
```
sirex, line 29: Would be nice to have a default location set, preferably using XDG_CACHE_HOME (~/.cache/vitrina/dist.db).

https://wiki.archlinux.org/title/XDG_Base_Directory

sirex (approved):

PR #736, opened 2023-11-09
347 automatinis vertimas

### LINKED ISSUE #347: Automatinis vertimas
Turi būti galimybė automatiškai išversti visus daugiakalbius [django-parler](https://github.com/atviriduomenys/katalogas/issues/61) pažymėtus verčiamus laukus.

Vertimas turėtu būti daromas naudojant https://vertimas.vu.lt/ automatinio vertimo sistemą, per API.
~Vertimas turėtu būti daromas naudojant http://vertimas.vdu.lt/twsas/PlugIns/Setup.aspx vertimo sistemą.~

Automatinis vertimas daromas iš Lietuvių į Anglų kalbą.

Tai universaliai, teoriškai turėtų būti galima įgyvendinti naudojant [Signalus](https://docs.djangoproject.com/en/3.2/topics/signals/):

https://django-parler.readthedocs.io/en/stable/api/parler.signals.html

### Epic

- https://github.com/atviriduomenys/katalogas/issues/17

### Susiję

- https://github.com/atviriduomenys/katalogas/issues/268

adp-atea:


vitrina/datasets/models.py
```diff
@@ -467,6 +477,49 @@ def get_download_count(self):
...
+                        "appId": "",
+                        "systemID": "smt-8abc06a7-09dc-405c-bd29-580edc74eb05",
+                        "text": lt_title,
+                        "options": ""
+                    },
+                    headers={
+                        "client-id": TRANSLATION_CLIENT_ID,
+                        "Content-Type": "application/json; charset=utf-8"
+                    },
+                )
```
sirex, line 501: It would be nice, if translation would be disabled if TRA.SLATION_CLIENT_ID is not given.

Also, it would be nice if form data saving would work even if VU service is down or misbehaving, for that we need to set [timeout to max 1 second](https://docs.python-requests.org/en/latest/user/advanced/#timeouts) and [handle possible request errors](https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions), probably we should handle all `RequestException` errors, and fail silently, adding just a log message about the error.

Also, since here we do two requests, maybe it would be better to do in [under single session](https://docs.python-requests.org/en/latest/user/advanced/#session-objects), to make use of Keep Alive and improve performance and cause less stress on VU servers?

sirex: Created a separate task for this:

- https://github.com/atviriduomenys/katalogas/issues/770

PR #1618, opened 2025-05-27
Generate and fill translations

Struchid:
Translations we're not generated for a long long time. Separating this PR to fill-out the missing translations, since currently, these are blocking colleagues, that are trying to add new changes, which require translated values.

Related pull request:
- https://github.com/atviriduomenys/katalogas/pull/1578


Struchid:
Already merged the bigger part.

PR #1653, opened 2025-06-30
[1603] Add missing migrations + transaction + translations

Struchid:
Bugfixes and adjustments:

- After moving the Agent to UAPI app, migrations were not commited, adding them now.
- Translations were added
- Added a transaction on agent creation, either all objects are created, or the transaction is roll-backed and the database will not have orphaned objects.


lescijuscct (approved):

martynasracct (approved):
LGTM

PR #1728, opened 2025-07-28
1710 UDTS endpoint to mark agent sync done and few agreement fixes

martynasracct:
**Summary:**
  - Add new endpoint `uapi/datasets/<form>/<org>/<catalog>/<catalog_sub>/Agreement/<agreement_id>/sync-done/` that changes `Agreement.status` to `ACTIVE` and updates `Agreement.last_sync_date`
  - Enable Agreement.is_agent_sync_enabled flag when agreement is signed
  - Allow creating project agreements for organizations that does not have one even if some agreements already exist
  - Make file paths relative in smart_contract tests

**Ticket:**
https://github.com/atviriduomenys/katalogas/issues/1710


vitrina/uapi/views/views.py
```diff
@@ -272,3 +275,30 @@ def list(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+
+        return queryset
+
+    def update(self, request: Request, *args: Any, **kwargs: Any) -> Response:
+        agreement = self.get_object()
+        agreement.last_sync_date = timezone.now()
+        agreement.status = AgreementStatuses.ACTIVE
+
+        # Agreement.updated_at must not be updated
+        agreement.save(update_fields=["last_sync_date", "status"])
```
varzgalys, line 302:
```suggestion
        agreement.save(update_fields=["last_sync_date", "status", "updated_at"])
```

varzgalys: If remember correctly - it still need to mentioned explicitly. (you confirm/decline with tests)

varzgalys (approved):

vitrina/uapi/views/views.py
martynasracct: You are correct. But I specifically need to not update `updated_at` field. That's why it's not included. I also left a comment about it

martynasracct:
Closing PR for now. Code from other branches has to be fixed before this PR

PR #1757, opened 2025-08-06
[1713] Change URL for credentials.cfg in Agents

Struchid:
Changing the URL, since it currently points to an old documentation not related to the Agent implementation.


martynasracct (approved):
LGTM 🔥

PR #1846, opened 2025-08-21
Moved fields from ResourceForm to DatasetResourceForm

Gustascct:
The `temporal_resolution` and `spatial_resolution` fields were in the `ResourceForm`, because there was no `DatasetResourceForm` but now `DatasetResourceForm` exists so the fields are moved to it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,13 +2,13 @@ Changes
 #######
 2025-08-21
 ==========
-https://github.com/atviriduomenys/katalogas/pull/1846
+https://github.com/atviriduomenys/katalogas/issues/642
 Moves `temporal_resolution` and `spatial_resolution` fields to `DatasetResourceForm` form.
 
 
 2025-08-21
 ==========
-https://github.com/atviriduomenys/katalogas/pull/1827
+https://github.com/atviriduomenys/katalogas/issues/642
 Adds `temporal_resolution` and `spatial_resolution` fields to `Dataset` model.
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -1,11 +1,17 @@
 Changes
 #######
+2025-08-21
+==========
+https://github.com/atviriduomenys/katalogas/pull/1846
```
Struchid: This URL points to this pull request, but it should point to the ticket w/ context instead.

CHANGES.rst
Struchid, line 1: Missing ticket in the description or relation to the ticket at the bottom right of the page.

Struchid (approved):

CHANGES.rst
Gustascct: Should the ticket be for the original problem this is solving or a new ticket created for this bug?

Struchid: If you have the new ticket, it would be best to use that. If not — use the original one.

Gustascct: But all urls in changes.rst are to the pull requests, not the original tickets, so this would create a mismatch.

martynasracct (approved):

CHANGES.rst
Struchid: Yeah, not your fault, leave it as is if you want, I'll raise the question and we will fix it. But since these are urgent, you can merge it without this change.

Struchid (approved):

PR #1852, opened 2025-08-25
Add missing dependencies

kastger:


karina-klinkeviciute (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.11-slim-bookworm AS builder
+FROM python:3.11-slim AS builder
```
varzgalys, line 1: sure about removing bookworm? I thing we should lock to more concrete variant if possible.

varzgalys (approved):

Dockerfile
kastger: This Dockerfile worked last time, I will merge this one now, for the sake of deploying and will test out if more specific option works too. Will raise new PR if it does.

PR #1877, opened 2025-08-29
1678: Adding try except case when no column is in DIM dict

MindaugasStanionis:


vitrina/datasets/structure.py
```diff
@@ -896,7 +900,11 @@ def _split_dim(
     lower = []
     found = None
     for dim in dims:
-        if row[dim]:
+        try:
+            value = row[dim]
+        except (KeyError, IndexError):
+            raise ValueError(dim)
```
varzgalys, line 906: I suggest creating custom, more explicit exception for this case. like MissingColumnException

varzgalys (approved):

tests/datasets/test_structure.py
```diff
@@ -132,3 +132,21 @@ def test_read_structure_table_not_mandatory_columns():
 
     assert props['id'].type == 'integer'
     assert len(props['description'].comments) == 1
+
+
+def test_read_structure_table_without_mandatory_columns_throws_error():
+    manifest_without_mandatory_columns = """\
```
Struchid, line 138: Is this needed?
```suggestion
    manifest_without_mandatory_columns = """
```

vitrina/datasets/structure.py
Struchid: I agree, we should add the `exceptions.py` file w/ some kind of a unique exception for this case.

Struchid (approved):

PR #1880, opened 2025-09-02
[1415] Adjustments & Bug fixes for the synchronization process

Struchid:
### Context:

Some additional changes had to be made to the implementation of the synchronization process:

1. Added API fields:
   - `service` & `series`: Deprecated but still used attributes, that need to be sent via the API to not break any existing logic.
   
   - `parent_id`: Defines relationships and dependencies of resources (Dataset model).

2. Added logic for `parent_id` setting, if the value is not sent — the object is set to root and has no children.
3. Found a bug:
   - When an object is initially saved via the `Dataset.add_root(instance=instance)`, the translations are not set, following that, ElasticSearch does not index the object and it is not displayed in the list of organizations. To solve that, we explicitly add translations & save the instance.

### Notes:
- Can be reviewed, tests will be added ASAP (in the same PR).

### Dependencies:
- https://github.com/atviriduomenys/spinta/pull/1450

### DIFF
--- a/vitrina/uapi/views/views.py
+++ b/vitrina/uapi/views/views.py
@@ -120,8 +120,26 @@ class DatasetViewSet(UAPIExceptionHandlerMixin, viewsets.ModelViewSet):
                 "access_rights": Dataset.NON_PUBLIC,
                 "organization_id": organization.id,
             }
+
+            title = dataset_data.get("title", None)
+            description = dataset_data.get("description", None)
+            parent_id = dataset_data.pop("parent_id", None)
+
             instance = Dataset(**dataset_data)
-            Dataset.add_root(instance=instance)
+
+            if parent_id:
+                parent = Dataset.objects.filter(id=parent_id).first()
+                parent.add_child(instance=instance)
+            else:
+                Dataset.add_root(instance=instance)
+
+            language = getattr(request, "LANGUAGE_CODE", "lt")
+            instance.set_current_language(language)
+            if title is not None:
+                instance.title = title
+            if description is not None:
+                instance.description = description
+            instance.save()
 
         Metadata.objects.create(
             uuid=str(uuid.uuid4()),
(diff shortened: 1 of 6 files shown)


vitrina/uapi/views/views.py
```diff
@@ -120,17 +120,35 @@ def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+
+            title = dataset_data.get("title", None)
+            description = dataset_data.get("description", None)
+            parent_id = dataset_data.pop("parent_id", None)
+
             instance = Dataset(**dataset_data)
-            Dataset.add_root(instance=instance)
+
+            if parent_id:
+                parent = Dataset.objects.filter(id=parent_id).first()
```
wisecrow: What if such Dataset record is missing? Maybe we should rise 404 error ?

wisecrow (approved):

vitrina/uapi/views/views.py
Struchid: Good catch, rewrote it in a bit of a different way:
- [5d0bc14](https://github.com/atviriduomenys/katalogas/pull/1880/commits/5d0bc143b9256abbb573e44f6c6559ff6a9d6afe)

Thanks!

vitrina/uapi/views/views.py
```diff
@@ -120,17 +120,35 @@ def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
+                parent.add_child(instance=instance)
+            else:
+                Dataset.add_root(instance=instance)
+
+            language = getattr(request, "LANGUAGE_CODE", "lt")
+            instance.set_current_language(language)
+            if title is not None:
+                instance.title = title
+            if description is not None:
+                instance.description = description
```
varzgalys: Optional:

```suggestion
            instance.title = title or instance.title
            instance.description = description or instance.description
```

varzgalys (approved):

vitrina/uapi/views/views.py
Struchid: Fixed via: [b3eb080](https://github.com/atviriduomenys/katalogas/pull/1880/commits/b3eb080e88d97aee5c280246d6d7ef55d7e623c8)

PR #1895, opened 2025-09-08
Create .drone-dev.yml

karina-klinkeviciute:
settings for deployment to a `dev` environment, which will be a new environment, before staging (currently `test`) environment.


LaurynasVenck (commented):
approved

PR #2014, opened 2025-10-15
1958 fix multi input field from adding multiple fields on single click

### LINKED ISSUE #1958: Multiple fields added on single click for FE multi insert
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/1399
#

When user click "+" sign to add new field, the amount of fields is multiplied by the number of times this field type is used on a form. 

Image

Problem is with field template `vitrina/templates/component/multi_input.html`, event listener is added each time form field is used, thus binding same even multiple times to each form element.

gts446:
Updated multi_input.html template javascript to trigger event listener only once per page.

### DIFF
--- a/vitrina/templates/component/multi_input.html
+++ b/vitrina/templates/component/multi_input.html
@@ -36,6 +36,8 @@
 
 <script>
 (function(){
+  if (window.__stringListBound) return;
+  window.__stringListBound = true;
   function updateRemoveButtons(container){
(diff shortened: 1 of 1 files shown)


martynasracct (approved):
Please, update pull request description. Otherwise, LGTM

Struchid (approved):

PR #2114, opened 2025-11-17
[1955]: fix Contact model migrations

wisecrow:
1. Remove duplicate contacts by email and organization
2. Make Contact.email unique together with organization


vitrina/datasets/models.py
```diff
@@ -1840,15 +1840,20 @@ class Contact(models.Model):
...
+    email = models.EmailField(_("Email"), blank=True)
     phone = models.CharField(_("Phone"), max_length=50, blank=True)
 
     class Meta:
         verbose_name = _("Kontaktas")
         verbose_name_plural = _("Kontaktai")
+        unique_together = (
+            "email",
+            "organization",
+        )
```
Struchid: Let's use unique_constraint instead. According to the documentation:

> Use [UniqueConstraint](https://docs.djangoproject.com/en/5.2/ref/models/constraints/#django.db.models.UniqueConstraint) with the [constraints](https://docs.djangoproject.com/en/5.2/ref/models/options/#django.db.models.Options.constraints) option instead.
[UniqueConstraint](https://docs.djangoproject.com/en/5.2/ref/models/constraints/#django.db.models.UniqueConstraint) provides more functionality than unique_together. unique_together may be deprecated in the future.
- https://docs.djangoproject.com/en/5.2/ref/models/options/#unique-together

Struchid (approved):

gts446:
Is it good idea to modify migrations, even though not yet in prod? 🤔

vitrina/datasets/models.py
wisecrow: Added. Thanks

wisecrow:
> Is it good idea to modify migrations, even though not yet in prod? 🤔

Assuming migrations for some people failed due to duplicates,  you have to modify existing migrations to apply it for failed cases.

martynasracct (approved):

PR #2115, opened 2025-11-18
Show all metadata in the publishing form

### LINKED ISSUE #2111: DSA struktūros publikavimo formoje nerodomi visi laukai
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/2098
#

### Bugas:
Duomenų rinkinio struktūros publikavimo formoje, "Įtraukiama į versiją" skiltyje nėra rodomi visi metaduomenys.

#### Tikrai nerodomi:
1) Prefix
2) Param laukai.
3) Dataset Distribution
4) Base (greičiausiai nereikia)

#### Rodomi nepilnai:
1) EnumItem. EnumItem rodomi, jei jie prijungti prie Property, jei jie prijungti prie Dataset, jie rodomi nėra. 
2) Property. Property pagaminti per link_denorm_prop funkciją nėra vaizduojami. Įtariu, kad taip yra dėl to, nes jie prie given stulpelio neturi 1. Šitie property yra sukuriami be naudotojo žinios.

### Tikslas:

Visi laukai esantys `Metadata` lentoje turi būti atvaizduoti Publikavimo formoje. Publikavimo formoje turi būti galima pasirinkti tiek laukų, kiek jų yra `Metadata` lentoje.

### Siūlomo sprendimo nuotrauka:

Image

Gustascct:
The publishing form was not displaying all of the metadata fields that were required. This made it difficult to collect the full version. 
This PR makes the publishing form hold all of the metadata fields as possible selections.

### DIFF
--- a/vitrina/datasets/models.py
+++ b/vitrina/datasets/models.py
@@ -989,10 +989,15 @@ class Dataset(Resource):
             content_type_id = metadata_instance.content_type.id
             metadata_model = ContentType.objects.get_for_id(content_type_id).model
             if metadata_model == "prefix":
+                prefix_text = (
+                    f"<a href='{metadata_instance.uri_link}'>Prefix</a>"
+                    if metadata_instance.uri_link else "Prefix"
+                )
                 label = mark_safe(
-                    f"<a href={metadata_instance.uri_link}>Prefix</a> name: "
+                    f"{prefix_text} name: "
                     f"<span class='tag is-success is-light is-medium'>{metadata_instance.name}</span>"
                 )
+
                 meta_objects.append((metadata_instance.pk, label))
 
             if metadata_model == "paramitem":
@@ -1001,8 +1006,12 @@ class Dataset(Resource):
                 if param.content_type_id == ContentType.objects.filter(model="dataset").first().pk and metadata_instance.metadata_version.status == VersionStatus.DRAFT:
                     if metadata_instance.ref:
                         dataset_param_item_metadata_ref = metadata_instance.ref
+                    param_text = (
+                        f"<a href='{metadata_instance.uri_link}'>Param</a>: {dataset_param_item_metadata_ref}"
+                        if metadata_instance.uri_link else f"Param: {dataset_param_item_metadata_ref}"
+                    )
                     label = mark_safe(
-                        f"Param: {dataset_param_item_metadata_ref} "
+                        f"{param_text} "
                         f"<span class='tag is-success is-light is-medium'>{metadata_instance.prepare}</span>"
                     )
                     meta_objects.append((metadata_instance.pk, label))
@@ -1020,25 +1029,12 @@ class Dataset(Resource):
                     meta_objects.append((metadata_instance.pk, label))
 
         for model in self.model_set.all():
-            if param := model.params.first():
-                for param_item in param.paramitem_set.all():
-                    metadata = param_item.metadata.first()
-                    if metadata and metadata.metadata_version.status == VersionStatus.DRAFT:
-                        if metadata.ref:
-                            model_param_item_metadata_ref = metadata.ref
-
-                        label = mark_safe(
-                            f"Param: <span class='prop_metadata'>{model_param_item_metadata_ref}</span> "
-                            f"<span class='tag is-success is-light is-medium'>{metadata.prepare}</span>"
-                        )
-                        meta_objects.append((metadata.pk, label))
-
             if dataset_distribution_for_model := model.distribution:
                 if dataset_distribution_metadata := dataset_distribution_for_model.metadata.first():
                     if dataset_distribution_metadata.metadata_version.status == VersionStatus.DRAFT:
                         label = mark_safe(
-                            f"<a href={dataset_distribution_for_model.get_absolute_url()}>{dataset_distribution_for_model.title}</a> name: "
-                            f"<span class='tag is-success is-light is-medium'>{dataset_distribution_for_model.title}</span>"
+                            f"<a href={dataset_distribution_for_model.get_absolute_url()}>{dataset_distribution_metadata.name}</a> name: "
+                            f"<span class='tag is-success is-light is-medium'>{dataset_distribution_metadata.name}</span>"
                         )
 
                         if dataset_distribution_metadata.pk not in dataset_distributions:
@@ -1102,6 +1098,25 @@ class Dataset(Resource):
                         )
                     label = mark_safe(label_str)
                     meta_objects.append((metadata.pk, label))
+
+                if param := model.params.first():
+                    for param_item in param.paramitem_set.all():
+                        metadata = param_item.metadata.first()
+                        if metadata and metadata.metadata_version.status == VersionStatus.DRAFT:
+                            if metadata.ref:
+                                model_param_item_metadata_ref = metadata.ref
+
+                            param_text = (
+                                f"<a href='{metadata.uri_link}'>Param</a>: "
+                                f"<span class='prop_metadata'>{model_param_item_metadata_ref}</span>"
+                                if metadata.uri_link else
+                                f"Param: <span class='prop_metadata'>{model_param_item_metadata_ref}</span>"
+                            )
+                            label = mark_safe(
+                                f"{param_text} "
+                                f"<span class='tag is-success is-light is-medium'>{metadata.prepare}</span>"
+                            )
+                            meta_objects.append((metadata.pk, label))
 
             for prop in model.model_properties.all():
                 metadata = prop.metadata.first()
--- a/vitrina/datasets/models.py
+++ b/vitrina/datasets/models.py
@@ -990,8 +990,7 @@ class Dataset(Resource):
             metadata_model = ContentType.objects.get_for_id(content_type_id).model
             if metadata_model == "prefix":
                 prefix_text = (
-                    f"<a href='{metadata_instance.uri_link}'>Prefix</a>"
-                    if metadata_instance.uri_link else "Prefix"
+                    f"<a href='{metadata_instance.uri_link}'>Prefix</a>" if metadata_instance.uri_link else "Prefix"
                 )
                 label = mark_safe(
                     f"{prefix_text} name: "
@@ -1003,12 +1002,16 @@ class Dataset(Resource):
             if metadata_model == "paramitem":
                 paramitem_instance = ParamItem.objects.filter(metadata=metadata_instance).first()
                 param = paramitem_instance.param
-                if param.content_type_id == ContentType.objects.filter(model="dataset").first().pk and metadata_instance.metadata_version.status == VersionStatus.DRAFT:
+                if (
+                    param.content_type_id == ContentType.objects.filter(model="dataset").first().pk
+                    and metadata_instance.metadata_version.status == VersionStatus.DRAFT
+                ):
                     if metadata_instance.ref:
                         dataset_param_item_metadata_ref = metadata_instance.ref
                     param_text = (
                         f"<a href='{metadata_instance.uri_link}'>Param</a>: {dataset_param_item_metadata_ref}"
-                        if metadata_instance.uri_link else f"Param: {dataset_param_item_metadata_ref}"
+                        if metadata_instance.uri_link
+                        else f"Param: {dataset_param_item_metadata_ref}"
                     )
                     label = mark_safe(
                         f"{param_text} "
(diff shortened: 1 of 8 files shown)


vitrina/datasets/models.py
```diff
@@ -1036,7 +1097,19 @@ def get_metadata_objects_for_version(self):
                     label = mark_safe(label_str)
                     meta_objects.append((metadata.pk, label))
 
-            for prop in model.model_properties.filter(given=True):
+                if param := model.params.first():
+                    for param_item in param.paramitem_set.all():
+                        metadata = param_item.metadata.first()
+                        if metadata and metadata.metadata_version.status == VersionStatus.DRAFT:
+                            if metadata.ref:
```
Struchid: These two can be combined to avoid level 4 nesting

vitrina/datasets/models.py
```diff
@@ -977,7 +984,61 @@ def get_metadata_objects_for_version(self):
...
+                    )
+                    meta_objects.append((metadata_instance.pk, label))
+
         for model in self.model_set.all():
+            if dataset_distribution_for_model := model.distribution:
+                if dataset_distribution_metadata := dataset_distribution_for_model.metadata.first():
+                    if (
+                        dataset_distribution_metadata.pk not in dataset_distributions
+                        and dataset_distribution_metadata.metadata_version.status == VersionStatus.DRAFT
+                    ):
```
Struchid: I suggest moving these checks to separate variables (regular value assignments to the walrus ones) and avoiding the indentation:


```suggestion
            dataset_distribution_for_model = model.distribution
            dataset_distribution_metadata = dataset_distribution_for_model.metadata.first()
            is_metadata_inside_expected_distributions = dataset_distribution_metadata.pk in dataset_distributions
            is_version_draft = dataset_distribution_metadata.metadata_version.status == VersionStatus.DRAFT
            
            if dataset_distribution_for_model and dataset_distribution_metadata and (not is_metadata_inside_expected_distributions and is_version_draft):
```

It will still be multiple lines of the conditional (because of line length) but we can easily avoid the nesting introduced here.

vitrina/datasets/models.py
```diff
@@ -977,7 +984,61 @@ def get_metadata_objects_for_version(self):
                 )
                 meta_objects.append((metadata.pk, label))
 
+        for metadata_instance in all_metadata_instances:
+            content_type_id = metadata_instance.content_type.id
+            metadata_model = ContentType.objects.get_for_id(content_type_id).model
```
Struchid: I suggest adding (and checking if it works as expected) 

`django.utils.functional import cached_property`

There are examples of its usage throughout Katalogas.

This should cache the result for the sent `content_type_id` and we avoid unnecessary database calls in this situation, and there is no reason for ContentType objects (and their ids) to change, so no problems with cached values should arise as well.

vitrina/datasets/models.py
```diff
@@ -977,7 +984,61 @@ def get_metadata_objects_for_version(self):
...
+            if metadata_model == "prefix":
+                label = mark_safe(
+                    f"Prefix ref: "
+                    f"<span class='tag is-success is-light is-medium prop_metadata'>{metadata_instance.name}</span>"
+                )
+
+                meta_objects.append((metadata_instance.pk, label))
+
+            if metadata_model == "paramitem":
+                paramitem_instance = ParamItem.objects.filter(metadata=metadata_instance).first()
```
Struchid:
```suggestion
                param_item_instance = ParamItem.objects.filter(metadata=metadata_instance).first()
```

vitrina/datasets/models.py
```diff
@@ -977,7 +984,61 @@ def get_metadata_objects_for_version(self):
...
+                        dataset_param_item_metadata_ref = metadata_instance.ref
+                    label = mark_safe(
+                        f"Param ref:  <span class='tag is-success is-light is-medium prop_metadata'>{dataset_param_item_metadata_ref}</span>"
+                        f" prepare: <span class='tag is-success is-light is-medium'>{metadata_instance.prepare}</span>"
+                    )
+                    meta_objects.append((metadata_instance.pk, label))
+
+            if metadata_model == "enumitem":
+                enumitem_instance = EnumItem.objects.filter(metadata=metadata_instance).first()
+                enum = enumitem_instance.enum
```
Struchid: Is it possible this is None? Further logic will fail if the answer is `yes`

vitrina/datasets/models.py
```diff
@@ -977,7 +984,61 @@ def get_metadata_objects_for_version(self):
...
+                    f"<span class='tag is-success is-light is-medium prop_metadata'>{metadata_instance.name}</span>"
+                )
+
+                meta_objects.append((metadata_instance.pk, label))
+
+            if metadata_model == "paramitem":
+                paramitem_instance = ParamItem.objects.filter(metadata=metadata_instance).first()
+                param = paramitem_instance.param
+                if (
+                    param.content_type_id == ContentType.objects.filter(model="dataset").first().pk
```
Struchid: Use `get_for_model` to avoid problems later:

`ContentType.objects.get_for_model(model)`

due to:
- Returns more than one ContentType (due to dataset being defined in different apps)
- Get instead of filter & first in this situation, since you expect only one particular content type

Struchid (approved):

vitrina/datasets/models.py
martynasracct: Is there a reason to do `ContentType.objects.get_for_id(content_type_id)`? Can't you just do `metadata_model = metadata_instance.content_type.model`?

Or am I misunderstanding something?

vitrina/datasets/models.py
martynasracct: Nitpick: it would be cleaner to get `param_item_instance` useing reverse generic relation and prefetch_related to possibly avoid n+1 queries.

https://docs.djangoproject.com/en/5.2/ref/contrib/contenttypes/#reverse-generic-relations

It should look something like this:
```
# Define related_query_name on ParamItem model
metadata = GenericRelation("Metadata", related_query_name="param_items")

# Use prefetch_related on Metadata query
all_metadata_instances = Metadata.objects.filter(dataset=self.pk, draft=True).prefetch_related("param_items")
```
and then something like
```suggestion
                paramitem_instances = metadata_instance.param_items.all()
                param_item_instance = paramitem_instances[0] if paramitem_instances else None
```

vitrina/datasets/models.py
```diff
@@ -977,7 +984,61 @@ def get_metadata_objects_for_version(self):
...
+                    if metadata_instance.ref:
+                        dataset_param_item_metadata_ref = metadata_instance.ref
+                    label = mark_safe(
+                        f"Param ref:  <span class='tag is-success is-light is-medium prop_metadata'>{dataset_param_item_metadata_ref}</span>"
+                        f" prepare: <span class='tag is-success is-light is-medium'>{metadata_instance.prepare}</span>"
+                    )
+                    meta_objects.append((metadata_instance.pk, label))
+
+            if metadata_model == "enumitem":
+                enumitem_instance = EnumItem.objects.filter(metadata=metadata_instance).first()
```
martynasracct: Same possible n+1 query here

vitrina/datasets/models.py
```diff
@@ -1036,7 +1097,19 @@ def get_metadata_objects_for_version(self):
                     label = mark_safe(label_str)
                     meta_objects.append((metadata.pk, label))
 
-            for prop in model.model_properties.filter(given=True):
+                if param := model.params.first():
```
martynasracct, line 1118: another n+1 query

vitrina/datasets/models.py
```diff
@@ -1036,7 +1097,19 @@ def get_metadata_objects_for_version(self):
...
+                    for param_item in param.paramitem_set.all():
```
martynasracct, line 1119: and another one

vitrina/datasets/models.py
```diff
@@ -1036,7 +1097,19 @@ def get_metadata_objects_for_version(self):
...
+                        metadata = param_item.metadata.first()
```
martynasracct, line 1120: and another one

martynasracct (commented):
There are a lot of n+1 queries in this code, because everything happens inside for loop and `select_related()` / `prefetch_related()` is not used.

The more metadata instances with related object there will be - the slower page will get because DB queries will grow exponentially

vitrina/datasets/models.py
Gustascct: Combined

vitrina/datasets/models.py
Gustascct: Changed to what Martynas suggested given that get_for_id is not needed. I could change to cached_property if you deem it necessary

vitrina/datasets/models.py
Gustascct: changed, thanks

vitrina/datasets/models.py
Gustascct: Used get_for_model

vitrina/datasets/models.py
Gustascct: added if statements that would check if its None. But overall the answer is no, you just made me scared.

vitrina/datasets/models.py
Gustascct: **To keep track:** 

Discussed this with @martynasracct and found out, that adding related_query_name does not create the possibility to use prefetch_related. Prefetch_related only works if the model is directly connected with another model. The N+1 problem could be solved, but it would require more work and most likely would not yield **amazing** results. So optimized where could and left it

vitrina/datasets/models.py
```diff
@@ -977,7 +986,79 @@ def get_metadata_objects_for_version(self):
...
+                            )
+                            meta_objects.append((metadata_instance.pk, label))
+
         for model in self.model_set.all():
+            dataset_distribution_for_model = model.distribution
+            dataset_distribution_metadata = (
+                dataset_distribution_for_model.metadata.first()
+                if dataset_distribution_for_model and dataset_distribution_for_model.metadata.exists()
+                else None
+            )
```
varzgalys, line 1037:
```suggestion
            dataset_distribution_metadata = (
                dataset_distribution_for_model.metadata.first() if dataset_distribution_for_model else None
            )
```
No point to check if exists(), first() will return None if nothing found.

vitrina/datasets/models.py
```diff
@@ -977,7 +986,79 @@ def get_metadata_objects_for_version(self):
...
+                label = mark_safe(
+                    f"Prefix ref: "
+                    f"<span class='tag is-success is-light is-medium prop_metadata'>{metadata_instance.name}</span>"
+                )
+
+                meta_objects.append((metadata_instance.pk, label))
+
+            if metadata_model == "paramitem":
+                param_item_instance = (
+                    ParamItem.objects.filter(metadata=metadata_instance).select_related("param").first()
```
varzgalys, line 1001: prefetch_related or manual collection (of all related records) prefered.

vitrina/datasets/models.py
```diff
@@ -977,7 +986,79 @@ def get_metadata_objects_for_version(self):
...
+                            if metadata_instance.ref:
+                                dataset_param_item_metadata_ref = metadata_instance.ref
+                            label = mark_safe(
+                                f"Param ref:  <span class='tag is-success is-light is-medium prop_metadata'>{dataset_param_item_metadata_ref}</span>"
+                                f" prepare: <span class='tag is-success is-light is-medium'>{metadata_instance.prepare}</span>"
+                            )
+                            meta_objects.append((metadata_instance.pk, label))
+
+            if metadata_model == "enumitem":
+                enum_item_instance = EnumItem.objects.filter(metadata=metadata_instance).select_related("enum").first()
```
varzgalys, line 1018: prefetch_related or manual collection (of all related records) prefered.

varzgalys (approved):
Performance related comments only.

vitrina/datasets/models.py
Gustascct: Thanks!

vitrina/datasets/models.py
Gustascct: Martynas had the same suggestion but we found out, that prefetch_related does not work with the data structure we have. Manual collection would be an improvement, but in the same discussion we thought that that would probably be an unnecessary use of time right now. I have it noted, that a performance upgrade is possible and if you are okay with it, I would like to come back to it if we start noticing issues

varzgalys: > Martynas had the same suggestion but we found out, that prefetch_related does not work with the data structure we have. Manual collection would be an improvement, but in the same discussion we thought that that would probably be an unnecessary use of time right now. I have it noted, that a performance upgrade is possible and if you are okay with it, I would like to come back to it if we start noticing issues

Ok, can skip for now. Maybe #TODO comment? (Optional)

Gustascct: Added

PR #2118, opened 2025-11-18
fix fakeviisp companycode field type

gts446:
updated `FakeViispForm` `lt_company_code` field type, because there was issue when trying codes that start with zero, and zero(s) are stripped.

### DIFF
--- a/vitrina/viisp/forms.py
+++ b/vitrina/viisp/forms.py
@@ -16,11 +16,10 @@ PROXY_TYPE_CHOICES = [
 
 class FakeViispForm(forms.Form):
     email = forms.EmailField(label=_("El. paštas"), required=True)
-    lt_company_code = forms.IntegerField(
+    lt_company_code = forms.CharField(
         label=_("Įmonės kodas"),
         help_text=_("Įveskite Lietuvoje registruotos įmonės kodą."),
         required=False,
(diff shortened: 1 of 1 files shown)


Struchid (approved):

vitrina/viisp/forms.py
```diff
@@ -16,11 +16,10 @@
 
 class FakeViispForm(forms.Form):
     email = forms.EmailField(label=_("El. paštas"), required=True)
-    lt_company_code = forms.IntegerField(
+    lt_company_code = forms.CharField(
         label=_("Įmonės kodas"),
         help_text=_("Įveskite Lietuvoje registruotos įmonės kodą."),
         required=False,
-        min_value=1,
     )
```
Gustascct, line 23: min_value=1 was also deleted, is that okay? Now it will not enforce the code to be > 1. It will also not enforce numbers only, is it intended?

Gustascct (approved):

vitrina/viisp/forms.py
gts446: Charfield doesn't support min_value parameter. No numbrs is fine, this tool is for development only, just fixing this because it caused confusions when trying numbers that start with 0

PR #2119, opened 2025-11-18
[2067] Agreement negotiation: Part 2.2

Struchid:
Additional changes before introducing quite big changes to agreement negotiation forms.

- Add additional statuses for agreement
- Adjust Django template formatting
- Fix a representational bug in templates
- Add the template field as non-required field
- Sanitize md text fields


wisecrow (approved):

vitrina/smart_contracts/models.py
```diff
@@ -82,6 +82,14 @@ class Agreement(UUIDBaseModel):
         null=True,
         blank=True,
     )
+    template = models.ForeignKey(
+        "vitrina_smart_contracts.SmartContractTemplate",
+        on_delete=models.SET_NULL,
+        related_name="agreements",
+        verbose_name=_("Sutarties šablonas"),
+        null=True,
```
varzgalys, line 90: Maybe better to have it non-optional. Old ones can either removed or filled with `SmartContractTemplate.objects.first()`

varzgalys (approved):

vitrina/smart_contracts/models.py
varzgalys: Or can agreement exists but not be generated yet?

Gustascct (approved):

vitrina/smart_contracts/models.py
Struchid: Agreement objects can exist and they might never have a template (until they reach a certain status, where the provider representative chooses which template to use), so I will be leaving this as an optional attribute.

PR #2145, opened 2025-11-26
Vertimas vu fix devel

kastger:
- Instead of leaving the translation empty, reused the same non translated value so its not empty when languages is changed on the page.
- Moved translation code into separate function in `utils.py` for reusability.
- Modified tests to reflect newest changes.

### DIFF
--- a/vitrina/utils.py
+++ b/vitrina/utils.py
@@ -1,0 +1,41 @@
+import logging
+import requests
+from vitrina.settings import TRANSLATION_CLIENT_ID, TRANSLATION_URL, TRANSLATION_REQUEST_TIMEOUT
+
+
+logger = logging.getLogger()
+
+
+def translate_text(text: str, field_name: str = "") -> str | None:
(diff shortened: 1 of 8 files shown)


vitrina/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        )
+        response.raise_for_status()
+        return response.json()
+    except requests.exceptions.Timeout:
+        logger.warning(f"Translation timeout for {field_name}")
+        return text
+    except requests.exceptions.RequestException as e:
+        logger.warning(f"Translation failed for {field_name}: {e}")
+        return text
+    except (ValueError, KeyError) as e:
```
Struchid, line 36: I would suggest also having a last `except` block which would handle anything else, we might get something we are not hoping to catch here

Struchid (approved):
Thanks for the change 🙌

martynasracct (approved):

PR #2169, opened 2025-12-03
Test 2124 template test fail

kastger:
Testing the fix to the problem of failing tests when using `override_settings(DEBUG=...)`


kastger:
Meh

PR #2176, opened 2025-12-04
FakeVIISPCompleteLoginView returns 404 if called withDEBUG=False

martynasracct:
**Summary:**

Fixes failing VIISP login tests if running locally with `DEBUG=True`.
```
(test_fake_viisp_not_available_when_debug_false) Failed: DID NOT RAISE <class 'django.urls.exceptions.NoReverseMatch'>
```

- `FakeVIISPCompleteLoginView` URL is always added to urlpatterns.
- `FakeVIISPCompleteLoginView` returns HTTP 404 from the view itself if called with `DEBUG=False`.
- Removed `tests.viisp.urls_fake_viisp` file. Tests use `@override_settings(DEBUG=...)` instead.


gts446 (approved):
LGTM.

MindaugasStanionis (approved):

kastger (approved):

Gustascct (approved):

vitrina/viisp/views.py
```diff
@@ -154,6 +156,11 @@ class FakeVIISPCompleteLoginView(LoginView):
     form_class = FakeViispForm
     redirect_authenticated_user = True
 
+    def dispatch(self, request: WSGIRequest, *args, **kwargs) -> HttpResponseBase:
+        if settings.DEBUG:
+            return super().dispatch(request, *args, **kwargs)
+        raise Http404()
```
Struchid: **Optional and very nitpicky:** I'd do this the other way around. Reasons for that:
1) Easier to extend, if more logic is required to be added (e.g. a new check which also would throw an error)
2) Success case comes last (as in all API calls, validations/checks first, then a 2XX response)

Completely optional, choose to ignore if you think some other way.

```suggestion
        if not settings.DEBUG:
            raise Http404()
        return super().dispatch(request, *args, **kwargs)
```

Struchid (approved):

vitrina/viisp/views.py
martynasracct: Completely agree with you on this

martynasracct: Fixed in a96ddbfb35829ac0e6b62c9b10a99b2915b536b0

wisecrow (approved):

PR #2179, opened 2025-12-08
2172 record all katalogas changes

### LINKED ISSUE #2172: Fiksuoti visus kataloge atliekamus duomenų pokyčius
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/2188
#

Analizės užduotis: https://github.com/atviriduomenys/dvms/issues/328

### Rolė, tikslas, funkcionalumas

Reikia "by default" fiksuoti visus* kataloge atliekamus duomenų pakeitimus, kad vėliau juos būtų galima audituoti ir, esant poreikiui, atstatyti į senesnę būseną.

*- išskyrus tiesiogiai nurodytiems modeliams ar aplikacijoms

### Aprašymas

_No response_

### As-Is (jei taikoma)

Šiuo metu pokyčiai fiksuojami tik tiesiogiai nurodytiems modeliams, bei tiesiogiai nurodytiems veiksmams.

### To-Be (pagrindinis)

Fiksuoti visus pokyčius, kurie yra atliekami kataloge:

- Per katalogo UI
  - Išskyrus `GET, OPTIONS, HEAD` HTTP metodus
- Per katalogo django admin
  - Išskyrus `GET, OPTIONS, HEAD` HTTP metodus
- Per Celery užduotis

Pokyčiai turėtų būti fiksuojami visiems modeliams, išskyrus tiesiogiai nurodytus, pvz:

```
NOT_VERSIONED_MODELS = (
    "vitrina.users.models.*",
    "vitrina.cms.models.CmsAttachment",
    ...
)
```

Atlikus bet kokį pokytį, papildomai būtų saugomas pokyčio kontekstas JSON formatu:

- **source**: `str`
  - `view` | `admin` | `celery`
- **action**: `str`
  - Priklauso nuo `source`.
    - Jeigu django admin, nurodomas veiksmas - `add` | `change` | `delete`
    - Jeigu per katalogo UI. Nurodomas `view`, Pvz: `vitrina.projects.views.ProjectUpdateView`
    - Jeigu celery task'as, nurodomas funkcijos/task'o pavadinimas. Pvz: `vitrina.structure.tasks.validate_manifest_task`
- **http_method**: `str`
  - Tik katalogo UI ir django admin. HTTP metodas, kurio metu buvo atliekamas pakeitimas.
  - pvz: `POST`
- **path**: `str`
  - Tik katalogo UI ir django admin
  - `/projects/485/change/`
  - `/admin/vitrina_projects/project/485/change/`
- **args**: `list`
  - `["485", "13748"]`
- **kwargs**: `dict`
  - `{"pk": 485}`
 
Pridėti prie django admin `reversion_revision` ir `reversion_version` lentas, kad būtų galima matyti pakeitimų įrašus.

### User Flow (jei taikoma)

_No response_

### Wireframes

_No response_

### Acceptance Criteria

Atlikus bet kokį pokyti, per django admin turi matytis naujas istorijos įrašas su pakeistų modelių snapshot'ais.

### Priedai

_No response_

### 📋 Definition of Done

- [x] **Testai** parašyti / atnaujinti
- [x] **Dokumentacija atnaujinta** (jei reikia)
- [x] Padeployinta ir ištestuota **Laikinojoj testinėj** aplinkoj
- [x] Acceptance Criteria įvykdyti
- [ ] CHANGES.rst
- [x] Pademonstruota per Sprint Review
- [x] PO (ar atsakingas asmuo) peržiūrėjo ir patvirtino

# Priėmimo testavimo scenarijus

### Testavimo eiga

----

1. Peržiūrima pakeitimų istorija per administravimo aplinką

----

### Testuojama kaip

Administratoriaus teises turintis naudotojas (VSSA darbuotojas)

#### Sąlygos

1. Duomenų rinkinys `test_duom` buvo sukurtas ir redaguotas institucijos tvarkytojo, naudojantis portalo naudotojo sąsaja
2. Duomenų rinkinys `test_duom` buvo redaguotas VSSA administratoriaus, naudojantis portalo administravimo sąsaja

#### Veiksmai

1. Atidaroma administravimo aplinka
2. Pasirenkama peržiūrėti pakeitimų įrašus

#### Įvedami duomenys

NA

#### Rezultatas

2.1. Pakeitimų įrašų lentelėje matomas `test_duom`
2.2. Matoma, kad `test_duom` buvo sukurtas ir redaguotas institucijos tvarkytojo
2.3. Matoma, kad `test_duom` buvo redaguotas VSSA administratoriaus
2.4. Matomi `test_duom` sukūrimo ir redagavimo laikai
2.5. Matomi `test_duom` pokyčių kontekstai

gts446:
**Summary**

This change enables global change/version tracking for katalogas using django-reversion, without requiring per-model registration or manual revision context management.

All changes are now automatically versioned for:

- Katalogas UI & DRF endpoints
- Django admin
- Celery tasks

Each revision also gets a JSON comment describing the request/task context to make it easier to trace where and how a change was made.

I suggest to review code by commits:

- `add reversion middleware for tracking changes`
- `add revision and version to django admin`
- `register all models with django reversion on project layer`
  - All project models are registered automatically unless listed in settings.NOT_VERSIONED_MODELS. Validation has been added to check if there are any registered models(manually with decorators or `VersionAdmin`) that are excluded.
- `replace existing bulk_create with create`
  - because bulk operations are not recorded by reversion.
- `track celery task changes with reversion`
- `add default revision comment in JSON for changes through django views` (and DRF endpoints)
- `add custom VersionAdmin class and update inherited model admin classes`
- `add JSON comment to revisions created by celery tasks`
  - Since it is possible to link user to task, but not possible to get user inside a task, added a "reserved" kwarg `user_id`, so user can be passed explicitly.
 - `remove manual revision comments`
  - Can be skipped.
 - `update broken tests`
  - Can be skipped.

TBD:
changes.rst, translations.


vitrina/admin.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+    readonly_fields = fields
+
+    def has_add_permission(self, request: HttpRequest, obj: Optional[Revision] = None) -> bool:
+        return False
+
+    def has_change_permission(self, request: HttpRequest, obj: Optional[Revision] = None) -> bool:
+        return False
+
+    @admin.display(description="Versijos duomenys")
```
martynasracct:
```suggestion
    @admin.display(description=_("Versijos duomenys"))
```

vitrina/admin.py
```diff
@@ -0,0 +1,158 @@
...
+                '<div id="{}" style="display:none; margin-top:4px;">'
+                "<pre style='white-space:pre-wrap; margin:0;'>{}</pre>"
+                "</div>"
+            ),
+            element_id,
+            element_id,
+            escape(pretty),
+        )
+
+    @admin.display(description="Modelis")
```
martynasracct:
```suggestion
    @admin.display(description=_("Modelis"))
```

vitrina/admin.py
```diff
@@ -0,0 +1,158 @@
...
+    def col_content_type(self, obj: Version) -> Any:
+        return obj.content_type
+
+    @admin.display(description="Objekto ID")
```
martynasracct:
```suggestion
    @admin.display(description=_("Objekto ID"))
```

vitrina/admin.py
```diff
@@ -0,0 +1,158 @@
...
+            url = reverse(
+                f"admin:{obj.content_type.app_label}_{obj.content_type.model}_change",
+                args=[obj.object_id],
+            )
+        except NoReverseMatch:
+            return str(obj.object_id)
+
+        return format_html('<a href="{}">{}</a>', url, obj.object_id)
+
+    @admin.display(description="Objektas")
```
martynasracct:
```suggestion
    @admin.display(description=_("Objektas"))
```

tests/api/test_views.py
```diff
@@ -544,7 +546,15 @@ def test_create_dataset(app: DjangoTestApp):
     assert list(dataset.category.all()) == [category]
     assert dataset.organization == organization
     assert Version.objects.get_for_object(dataset).count() == 1
-    assert Version.objects.get_for_object(dataset).first().revision.comment == Dataset.CREATED
+    revision_comment = RevisionComment(
```
Struchid: This should be instantiated at the beginning of the test.

Tests are separated into three blocks, called AAA in short (Arrange, Act, Assert), this belongs to the arrange part, and should not be in between the asserts.

tests/api/test_views.py
```diff
@@ -612,13 +622,22 @@ def test_update_dataset_with_dataset_id(app: DjangoTestApp):
...
-    res = app.patch(reverse("api-single-dataset", kwargs={'datasetId': dataset.pk}), {
+    url = reverse("api-single-dataset", kwargs={'datasetId': dataset.pk})
+    res = app.patch(url, {
         'title': "Updated title",
         'description': "Updated description"
     })
     dataset.refresh_from_db()
     assert Version.objects.get_for_object(dataset).count() == 1
-    assert Version.objects.get_for_object(dataset).first().revision.comment == Dataset.EDITED
+    revision_comment = RevisionComment(
```
Struchid, line 627: Same for this and all other methods below

tests/api/test_views.py
```diff
@@ -758,16 +793,24 @@ def test_delete_dataset_with_internal_id(app: DjangoTestApp):
...
+        source=RevisionSource.VIEW,
+        action="api-single-dataset-internal",
+        http_method="DELETE",
+        path=url,
+        args=[],
+        kwargs=kwargs_dict
+    )
+    assert Version.objects.get_for_object(dataset).first().revision.comment == revision_comment.to_json()
     assert Version.objects.get_for_object(dataset).first().revision.user == representative.user
```
Struchid: Use variables to not repeat retrievals from db, e.g. of reoccurring parts:

- `Version.objects.get_for_object(dataset)`
- `Version.objects.get_for_object(dataset).first().revision`

tests/comments/test_views.py
Struchid, line 1: Same comments as for tests/api/test_views.py for this and all other `test_views.py` files

vitrina/__init__.py
martynasracct, line 1: I already saw this change in another PR

tests/test_apps.py
```diff
@@ -0,0 +1,44 @@
...
+from django.test import override_settings
+from vitrina.utils import get_all_models, is_model_versioned
+from django.contrib import admin
+from vitrina.admin import RevisionCommentVersionAdmin
+
+import reversion
+
+
+def test_reversion_all_models_registered_or_excluded():
+    for model in get_all_models():
```
Struchid: Use parametrization instead of looping through all models because:
a) Logic like loops should not be introduced in tests (I guess the conditional `if` statement is OK, unless you can filter the get_all_models() by versioned/unversioned, then two separate tests for registered/excluded would be even nicer.
b) This will fail on the first incorrect/unexpected record, if there are more of them, you will have to run the tests multiple times and fix the issues one-by-one, otherwise, you'd see a list of failed tests (this single test, but parametrized) and could fix all of the models in one go, without re-running the tests.

tests/test_apps.py
```diff
@@ -0,0 +1,44 @@
...
+
+    with pytest.raises(
+        ImproperlyConfigured, match="The NOT_VERSIONED_MODELS pattern 'notExistingModel' does not match any models."
+    ):
+        api_config.ready()
+
+
+@pytest.mark.django_db
+def test_versioned_models_inherit_revision_comment_admin_class():
+    for model in get_all_models():
```
Struchid: Same comment about looping in tests

vitrina/api_example/admin.py
```diff
@@ -1,4 +1,8 @@
 from django.contrib import admin
 from .models import ApiExample
+from vitrina.admin import RevisionCommentVersionAdmin
 
-admin.site.register(ApiExample)
+
+@admin.register(ApiExample)
+class ApiExampleAmin(RevisionCommentVersionAdmin):
```
Struchid:
```suggestion
class ApiExampleAdmin(RevisionCommentVersionAdmin):
```

vitrina/admin.py
```diff
@@ -0,0 +1,136 @@
...
+    @admin.display(description="Versijos duomenys")
+    def version_data(self, obj: Version) -> str:
+        data = obj.field_dict
+        pretty = json.dumps(data, indent=2, ensure_ascii=False, default=str)
+
+        element_id = f"snapshot-{obj.pk}"
+
+        return format_html(
+            (
+                '<a href="#" onclick="'
```
martynasracct, line 62: What does this HTML and why is it needed?

vitrina/admin.py
```diff
@@ -0,0 +1,136 @@
...
+    readonly_fields = fields
+
+    def has_add_permission(self, request: HttpRequest, obj: Optional[Revision] = None) -> bool:
+        return False
+
+    def has_change_permission(self, request: HttpRequest, obj: Optional[Revision] = None) -> bool:
+        return False
+
+    @admin.display(description="Versijos duomenys")
+    def version_data(self, obj: Version) -> str:
```
martynasracct, line 43: Do we really need to show data of every element in list view?

vitrina/datasets/models.py
```diff
@@ -1383,8 +1381,8 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
         existing_urls = set(ApplicableLegislation.objects.filter(url__in=urls).values_list("url", flat=True))
         new_urls = [url for url in urls if url not in existing_urls]
 
-        if new_urls:
-            ApplicableLegislation.objects.bulk_create([ApplicableLegislation(url=url) for url in new_urls])
+        for url in new_urls:
```
Struchid: You're moving from `bulk_create` due to reversion, but this is not enough then:
1) Let's collect the records being created, to not have to call the database to get all of them
2) Have you checked for timeouts? How many URL's do I need to add to the form, to crash the web with a timeout? Have you considered celery for this?

vitrina/datasets/models.py
```diff
@@ -1398,8 +1396,8 @@ def update_documentation(self, urls: list[str]) -> None:
         )
         new_urls = [url for url in urls if url not in existing_urls]
 
-        if new_urls:
-            Documentation.objects.bulk_create([Documentation(documentation_link=url) for url in new_urls])
+        for url in new_urls:
+            Documentation.objects.create(documentation_link=url)
 
         all_entries = Documentation.objects.filter(documentation_link__in=urls)
```
Struchid: Same comments

vitrina/orgs/admin.py
```diff
@@ -70,7 +70,7 @@ def has_delete_permission(self, request: HttpRequest, obj: object | None = None)
         return request.user.is_staff or request.user.is_superuser
 
 
-class OrganizationAdmin(VersionAdmin, TreeAdmin):
+class OrganizationAdmin(TreeAdmin, RevisionCommentVersionAdmin):
```
Struchid: I'd be careful mixing the base classes around and leave it as was:


```suggestion
class OrganizationAdmin(RevisionCommentVersionAdmin, TreeAdmin):
```

Unless you are completely sure, that no inheritance-based problems arise from that

vitrina/resources/models.py
```diff
@@ -384,8 +382,8 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
         existing_urls = set(ApplicableLegislation.objects.filter(url__in=urls).values_list("url", flat=True))
         new_urls = [url for url in urls if url not in existing_urls]
 
-        if new_urls:
-            ApplicableLegislation.objects.bulk_create([ApplicableLegislation(url=url) for url in new_urls])
+        for url in new_urls:
+            ApplicableLegislation.objects.create(url=url)
 
         all_entries = ApplicableLegislation.objects.filter(url__in=urls)
```
Struchid: Again, let's collect the entries instead of calling the db again

vitrina/settings.py
```diff
@@ -545,3 +547,10 @@
 OTP_EMAIL_TOKEN_VALIDITY = 600
 
 USE_OTP_VALIDATION = env.bool("USE_OTP_VALIDATION", default=True)
+
+# List model paths to exclude from versioning.
+# Supports exact model paths or wildcards.
+# Examples:
+#   "vitrina.users.models.*"          → excludes all models in that module
+#   "vitrina.comments.models.Comment" → excludes a specific model
+NOT_VERSIONED_MODELS = []
```
Struchid: Use a tuple/set instead, they're faster to access

vitrina/admin.py
```diff
@@ -0,0 +1,136 @@
...
+
+    fields = (
+        "version_data",
+        "col_content_type",
+        "col_object_id",
+        "col_object_repr",
+    )
+    readonly_fields = fields
+
+    def has_add_permission(self, request: HttpRequest, obj: Optional[Revision] = None) -> bool:
```
Struchid: Let's use built-in python types and not from Typing


```suggestion
    def has_add_permission(self, request: HttpRequest, obj: Revision | None = None) -> bool:
```

Same for other methods as well

vitrina/admin.py
```diff
@@ -0,0 +1,136 @@
...
+        return False
+
+    def has_change_permission(self, request: HttpRequest, obj: Optional[Revision] = None) -> bool:
+        return False
+
+    @admin.display(description="Versijos duomenys")
```
Struchid: This is not translatable?

vitrina/admin.py
```diff
@@ -0,0 +1,136 @@
...
+    def version_data(self, obj: Version) -> str:
+        data = obj.field_dict
+        pretty = json.dumps(data, indent=2, ensure_ascii=False, default=str)
```
Struchid: Variable names should not be adjectives, they should be pronouns

vitrina/admin.py
```diff
@@ -0,0 +1,136 @@
...
+                "<pre style='white-space:pre-wrap; margin:0;'>{}</pre>"
+                "</div>"
+            ),
+            element_id,
+            element_id,
+            escape(pretty),
+        )
+
+    @admin.display(description="Modelis")
+    def col_content_type(self, obj: Version) -> Any:
```
Struchid: Let's use full names where possible


```suggestion
    def column_content_type(self, obj: Version) -> Any:
```

vitrina/apps.py
```diff
@@ -1,6 +1,38 @@
...
+from django.db import models
+from django.conf import settings
+from vitrina.utils import get_all_models, is_model_versioned
 
 
 class ApiConfig(AppConfig):
     name = "vitrina"
     label = "vitrina"
+
+    def ready(self):
```
Struchid: The return typehint is missing

vitrina/apps.py
```diff
@@ -1,6 +1,38 @@
...
+        self._register_models_for_reversion(versioned_models)
+
+    def _register_models_for_reversion(self, versioned_models: set[type[models.Model]]) -> None:
+        try:
+            for model in versioned_models:
+                try:
+                    reversion.register(model)
+                except reversion.RegistrationError:
+                    # Already registered
+                    pass
```
Struchid: Why not do a check first instead of catching the exception?

vitrina/celery.py
```diff
@@ -1,10 +1,70 @@
...
 app.config_from_object("django.conf:settings", namespace="CELERY")
-app.autodiscover_tasks()
 
 app.conf.timezone = "UTC"
+
+
+class RevisionedTask(app.Task):
+    abstract = True
+
+    def apply_async(self, args=None, kwargs=None, **options):
```
Struchid: typehints are missing

vitrina/celery.py
```diff
@@ -1,10 +1,70 @@
...
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "vitrina.settings")
 app = Celery("vitrina")
 app.config_from_object("django.conf:settings", namespace="CELERY")
-app.autodiscover_tasks()
 
 app.conf.timezone = "UTC"
+
+
+class RevisionedTask(app.Task):
```
Struchid, line 16: Let's add a docstring explaining what this is, why is it needed, and how it does it

vitrina/celery.py
```diff
@@ -1,10 +1,70 @@
...
+
+            if user_id:
+                User = get_user_model()
+                try:
+                    user = User.objects.get(pk=user_id)
+                except User.DoesNotExist:
+                    user = None
+
+                if user:
+                    reversion.set_user(user)
```
Struchid, line 53: Who is the author of a change, if the author is not a specific user?

vitrina/celery.py
```diff
@@ -1,10 +1,70 @@
...
+                source=RevisionSource.TASK,
+                action=self.name,
+                args=list(args),
+                kwargs=kwargs,
+            )
+            reversion.set_comment(comment.to_json())
+            return super().__call__(*args, **kwargs)
+
+    def _accepts_kwarg(self, kwarg_name: str) -> bool:
+        signature = inspect.signature(self.run)
```
Struchid: We should cache this, it will not be different everytime, right?

vitrina/celery.py
```diff
@@ -1,10 +1,70 @@
...
+        for parameter in signature.parameters.values():
+            if parameter.kind == inspect.Parameter.VAR_KEYWORD:
+                return True
+            if parameter.name == kwarg_name:
+                return True
+        return False
```
Struchid: Could be simplified using `any`

vitrina/celery.py
Struchid, line 1: All functions should have docstrings here

vitrina/middleware.py
```diff
@@ -23,3 +25,33 @@ def __call__(self, request):
                 logout(request)
         response = self.get_response(request)
         return response
+
+
+class AutoRevisionCommentMiddleware(MiddlewareMixin):
+    """
+    Automatically sets a default reversion comment based on the view
```
Struchid: When a CUD operations is performed, right? Should be specified

vitrina/utils.py
```diff
@@ -39,3 +48,69 @@ def translate_text(text: str, field_name: str = "") -> str | None:
     except Exception as e:
         logger.exception(f"Unexpected error during translation for {field_name}: {e}")
         return text
+
+
+def is_model_versioned(model: type[models.Model]) -> bool:
+    if model not in get_all_models(app_prefix="vitrina"):
+        return False
+    excluded_models_patterns = settings.NOT_VERSIONED_MODELS
```
Struchid: Just use the constant without redefining it into a separate variable with a different naming, if this is done only because of the naming, might as well change the name of the constant.

vitrina/utils.py
Struchid, line 1: All of these need docstrings as well

vitrina/utils.py
```diff
@@ -39,3 +48,69 @@ def translate_text(text: str, field_name: str = "") -> str | None:
...
+
+def get_all_models(
+    app_prefix: str | None = None, include_proxy: bool = True, include_unmanaged: bool = False
+) -> set[type[models.Model]]:
+    project_models = set()
+    for app_config in apps.get_app_configs():
+        if app_prefix and not app_config.name.startswith(app_prefix):
+            continue
+
+        for model in app_config.get_models():
```
Struchid: Why is a loop in a loop needed here? `apps.get_models()` should be enough here, `_meta` attribute from model will help get the `app_label`

vitrina/admin.py
gts446: This is for django admin. Does it support multi-language in katalogas? 🤔

vitrina/celery.py
gts446: user is optional. So left blank if not specified.

tests/api/test_views.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/4d27a89c717272aaaa7d7ef04ebb085f78e3efac

tests/api/test_views.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/4d27a89c717272aaaa7d7ef04ebb085f78e3efac

tests/api/test_views.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/4d27a89c717272aaaa7d7ef04ebb085f78e3efac

tests/comments/test_views.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/4d27a89c717272aaaa7d7ef04ebb085f78e3efac

tests/test_apps.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/4d27a89c717272aaaa7d7ef04ebb085f78e3efac

tests/test_apps.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/4d27a89c717272aaaa7d7ef04ebb085f78e3efac

vitrina/api_example/admin.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/orgs/admin.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/settings.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/admin.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/admin.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/admin.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/apps.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/apps.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/celery.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/celery.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/47977a3c6845d3bab3e911617ad31954609103e4

vitrina/celery.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/447fd4af8020e9143139e7a5ee3e7526561d595e

vitrina/datasets/models.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/3f6eb828f53d6bb4756914726d248597f6014b7d

vitrina/datasets/models.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/3f6eb828f53d6bb4756914726d248597f6014b7d

vitrina/resources/models.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/3f6eb828f53d6bb4756914726d248597f6014b7d

vitrina/celery.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/9ba987bbfa32e14ba8c34b64c56735c2d4cf959d

vitrina/middleware.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/9ba987bbfa32e14ba8c34b64c56735c2d4cf959d

vitrina/utils.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/9ba987bbfa32e14ba8c34b64c56735c2d4cf959d

vitrina/utils.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/9ba987bbfa32e14ba8c34b64c56735c2d4cf959d

vitrina/utils.py
gts446: https://github.com/atviriduomenys/katalogas/pull/2179/commits/9ba987bbfa32e14ba8c34b64c56735c2d4cf959d

vitrina/admin.py
Struchid: It does, and let's continue the good practice (even though not all are translated):

image

image

vitrina/celery.py
```diff
@@ -1,10 +1,90 @@
...
+
+            if user_id:
+                User = get_user_model()
+                try:
+                    user = User.objects.get(pk=user_id)
+                except User.DoesNotExist:
+                    user = None
+
+                if user:
+                    reversion.set_user(user)
```
martynasracct:
```suggestion
            if request:
                headers = getattr(request, "headers", {}) or {}

            if user_id := headers.get("user_id"):
                User = get_user_model().
                try:
                    user = User.objects.get(pk=user_id)
                    reversion.set_user(user)
                except User.DoesNotExist:
                    user = None
```

vitrina/celery.py
```diff
@@ -1,10 +1,90 @@
...
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "vitrina.settings")
 app = Celery("vitrina")
 app.config_from_object("django.conf:settings", namespace="CELERY")
-app.autodiscover_tasks()
 
 app.conf.timezone = "UTC"
+
+
+class RevisionedTask(app.Task):
```
martynasracct, line 16: An example of how to use `RevisionedTask` would be appreciated. Will this work with `@celery_app.task` and `@shared_task` decorators?

Also we should cover this with tests

vitrina/middleware.py
```diff
@@ -23,3 +25,33 @@ def __call__(self, request):
...
         return response
+
+
+class AutoRevisionCommentMiddleware(MiddlewareMixin):
+    """
+    Automatically sets a default reversion comment for CUD operations based on the view
+    and the view arguments, if a revision is active.
+    """
+
+    def process_view(self, request, view_func, view_args, view_kwargs):
```
martynasracct: type hints are missing

vitrina/middleware.py
martynasracct, line 1: Tests for middleware are missing

vitrina/services.py
```diff
@@ -10,7 +10,6 @@
 
 
 def fetch_page_title(url: str, timeout: int = DEFAULT_FETCH_TIMEOUT) -> str | None:
```
martynasracct, line 12: There is no longer a need to move this to Celery task?

tests/test_apps.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@pytest.mark.parametrize(
+        'versioned_model',
+        [model for model in get_all_models() if is_model_versioned(model)]
+)
+@pytest.mark.django_db
+def test_versioned_models_inherit_revision_comment_admin_class(versioned_model: type[models.Model]):
+    model_admin_instance = admin.site._registry.get(versioned_model)
+    if model_admin_instance:
```
martynasracct:
```suggestion
    if model_admin_instance := admin.site._registry.get(versioned_model):
```

vitrina/classifiers/models.py
```diff
@@ -265,9 +262,3 @@ class Meta:
 
     def __str__(self) -> str:
         return self.description or self.url
-
-    def update_description(self) -> str | None:
```
martynasracct, line 269: Why is this no longer needed?

tests/resources/test_views.py
```diff
@@ -99,6 +101,49 @@ def test_add_form_correct_login(app: DjangoTestApp):
...
+    form['format'] = file_format.id
+    form['download_url'] = "www.google.lt"
+    form["applicable_legislation"] = applicable_legislation_urls
+    with patch("vitrina.datasets.tasks.update_applicable_legislation_description.delay") as mocked_task:
+            resp = form.submit()
+
+    assert resp.status_code == 302
+    assert mocked_task.call_count == 2
+    assert DatasetDistribution.objects.filter().count() == 1
+    dataset_distribution = DatasetDistribution.objects.first()
```
martynasracct:
```suggestion
    dataset_distribution = DatasetDistribution.objects.first()
    assert dataset_distribution
```

tests/resources/test_views.py
```diff
@@ -99,6 +101,49 @@ def test_add_form_correct_login(app: DjangoTestApp):
...
+        field.value = new_urls[i] if i < len(new_urls) else ""
+
+    with patch("vitrina.datasets.tasks.update_applicable_legislation_description.delay") as mocked_task:
+        resp = form.submit()
+        
+    resource.refresh_from_db()
+    assert resp.status_code == 302
+    assert mocked_task.call_count == 2
+    assert DatasetDistribution.objects.filter().count() == 1
+    dataset_distribution = DatasetDistribution.objects.first()
```
martynasracct:
```suggestion
    dataset_distribution = DatasetDistribution.objects.first()
    assert dataset_distribution
```

martynasracct (commented):
Since there are a lot of custom logic added that is not documented in usages of reversion package, we should document how to add/remove models from versioning. Same for admin and celery tasks

vitrina/admin.py
gts446: Some "quick and cheap" UX improvement when viewing versioned objects on django admin. It adds expandable view of serialized object snapshot.  Since it is not "for public" I decided we don't need anything fancy here. Do you agree?
image

martynasracct: If I were you, I wouldn't show this JSON in ListView at all

vitrina/admin.py
gts446: I thought better to provide as much info as possible, since its easy to access and for admins only. I could remove if you know why this is bad idea?

vitrina/celery.py
gts446: Removed this part completely.

vitrina/celery.py
gts446: Removed this part completely.
Updated kwarg name to be explicit (`_reversion_user_id`) for this purpose, as before this could cause issues if developer uses `user_id` for task logic and for reversion `user_id`,  because they might serve different purpose.

vitrina/services.py
gts446: I already did, I guess the comment wasn't really in a right place, it should have been inside `Model.update_applicable_legislation`

vitrina/classifiers/models.py
gts446: I moved this logic into celery task

vitrina/celery.py
gts446: This doesnt change usage of celery, a.k.a tasks should be called as usual. Only additional feature that is not default behavior is that when calling tasks with `task_name.delay(...)` custom optional argument `_revision_user_id` can be passed, which will be used as revision author/user. I will document this.
Added tests.

vitrina/admin.py
martynasracct: I don't see a lot of benefits of having it tbf. You can always open it and see all data.

vitrina/services.py
martynasracct: Yeah, I think I wrote this comment few days ago. You made changes after that

vitrina/admin.py
gts446: Discussed separately. Can be closed.

vitrina/admin.py
gts446: Discussed separately. Can be closed.

tests/test_apps.py
```diff
@@ -0,0 +1,54 @@
...
+        for model in get_all_models()
+    ],
+)
+def test_reversion_model_registration(model: type[models.Model], should_be_registered: bool):
+    path = f"{model.__module__}.{model.__name__}"
+
+    if should_be_registered:
+        assert reversion.is_registered(model), (
+            f"Model '{path}' should be registered with reversion."
+        )
```
Struchid: Line length is 120, should be more than enough to put in one line
```suggestion
        assert reversion.is_registered(model), f"Model '{path}' should be registered with reversion."
```

tests/test_apps.py
```diff
@@ -0,0 +1,54 @@
...
+        api_config.ready()
+
+
+@pytest.mark.parametrize(
+        'versioned_model',
+        [model for model in get_all_models() if is_model_versioned(model)]
+)
+@pytest.mark.django_db
+def test_versioned_models_inherit_revision_comment_admin_class(versioned_model: type[models.Model]):
+    if model_admin_instance:= admin.site._registry.get(versioned_model):
```
Struchid:
```suggestion
    if model_admin_instance := admin.site._registry.get(versioned_model):
```

tests/test_middleware.py
```diff
@@ -0,0 +1,70 @@
...
+
+from vitrina.middleware import AutoRevisionCommentMiddleware
+from vitrina.utils import RevisionComment, RevisionSource
+
+
+def test_noop_when_reversion_not_active(rf: RequestFactory):
+
+    middleware = AutoRevisionCommentMiddleware(get_response=lambda r: None)
+    with patch("vitrina.middleware.reversion.is_active", return_value=False), \
+        patch("vitrina.middleware.reversion.set_comment") as set_comment_mock:
```
Struchid: Use parantheses to avoid slash line-breaks
```suggestion
    with (
        patch("vitrina.middleware.reversion.is_active", return_value=False),
        patch("vitrina.middleware.reversion.set_comment") as set_comment_mock
    ):
```

tests/test_middleware.py
```diff
@@ -0,0 +1,70 @@
...
+        middleware.process_view(rf.post(""), view_func=lambda r: None, view_args=[], view_kwargs={})
+
+    set_comment_mock.assert_not_called()
+
+
+def test_noop_when_comment_already_set(rf: RequestFactory):
+
+    middleware = AutoRevisionCommentMiddleware(get_response=lambda r: None)
+    with patch("vitrina.middleware.reversion.is_active", return_value=True), \
```
Struchid: Same for here and other places in this file

vitrina/datasets/tasks.py
```diff
@@ -0,0 +1,12 @@
...
+from vitrina.services import fetch_page_title
+from vitrina.classifiers.models import ApplicableLegislation
+
+
+@shared_task
+def update_applicable_legislation_description(legislation_id: UUID) -> None:
+    legislation = ApplicableLegislation.objects.get(pk=legislation_id)
+    if title := fetch_page_title(legislation.url):
+        legislation.description = title
+        legislation.save(update_fields=["description"])
```
Struchid: `updated_at` does not update without explicitly adding it to the collection of `update_fields`
```suggestion
        legislation.save(update_fields=["description", "updated_at"])
```

vitrina/resources/models.py
```diff
@@ -381,14 +380,12 @@ def save_translations(self, *args, **kwargs):
...
-        existing_urls = set(ApplicableLegislation.objects.filter(url__in=urls).values_list("url", flat=True))
-        new_urls = [url for url in urls if url not in existing_urls]
+        legislations: list[ApplicableLegislation] = []
 
-        if new_urls:
-            ApplicableLegislation.objects.bulk_create([ApplicableLegislation(url=url) for url in new_urls])
+        for url in urls:
+            legislation, created = ApplicableLegislation.objects.get_or_create(url=url)
+            if created:
+                update_applicable_legislation_description.delay(legislation.uuid)
```
Struchid: Wait, but why do we need to issue an N of celery jobs here and not just collect legislation uuid's and start a single job which could call those urls individually inside the job. Batch them instead.

vitrina/resources/views.py
```diff
@@ -285,28 +281,25 @@ def handle_no_permission(self):
...
-                    object_id=dataset.pk,
-                    type=Comment.STATUS,
-                    status=comment_status,
-                    user=self.request.user,
-                )
-                dataset.save()
+        add_to_revision(resource)
+        resource.delete()
+
+        if not DatasetDistribution.objects.filter(dataset=dataset) and dataset.is_public:
```
Struchid: Quicker & less resource heavy check:

```suggestion
        if not DatasetDistribution.objects.filter(dataset=dataset).exists() and dataset.is_public:
```

vitrina/uapi/views/views.py
```diff
@@ -118,30 +117,29 @@ def create(self, request: Request, *args: Any, **kwargs: Any) -> Response:
...
 
-            instance = Dataset(**dataset_data)
+        instance = Dataset(**dataset_data)
 
-            parent = Dataset.objects.filter(id=parent_id).first() if parent_id else None
-            if parent:
-                parent.add_child(instance=instance)
-            else:
-                Dataset.add_root(instance=instance)
+        parent = Dataset.objects.filter(id=parent_id).first() if parent_id else None
```
Struchid:
```suggestion
        if parent_id and (parent := Dataset.objects.filter(id=parent_id).first()):
            parent.add_child(instance=instance)
        else:
            Dataset.add_root(instance=instance)
```

vitrina/admin.py
```diff
@@ -0,0 +1,137 @@
...
+        "column_object_id",
+        "column_object_repr",
+    )
+    readonly_fields = fields
+
+    def has_add_permission(self, request: HttpRequest, obj: Revision | None = None) -> bool:
+        return False
+
+    def has_change_permission(self, request: HttpRequest, obj: Revision | None = None) -> bool:
+        return False
```
Struchid, line 37: We should also add a `has_delete_permission` method then as well

vitrina/admin.py
```diff
@@ -0,0 +1,137 @@
...
+                "}}"
+                "return false;"
+                '">Rodyti &#9660;</a>'
+                '<div id="{}" style="display:none; margin-top:4px;">'
+                "<pre style='white-space:pre-wrap; margin:0;'>{}</pre>"
+                "</div>"
+            ),
+            element_id,
+            element_id,
+            escape(data_json),
```
Struchid: format_html should already escape arguments, so this `escape` is redundant

vitrina/apps.py
```diff
@@ -1,6 +1,35 @@
...
+        self._register_models_for_reversion(versioned_models)
+
+    def _register_models_for_reversion(self, versioned_models: set[type[models.Model]]) -> None:
+        try:
+            for model in versioned_models:
+                if not reversion.is_registered(model):
+                    reversion.register(model)
+
+        except (OperationalError, ProgrammingError):
+            pass
```
Struchid: Log the error here to keep track of unexpected errors in the system

vitrina/celery.py
```diff
@@ -1,10 +1,73 @@
...
+
+    Responsibilities:
+    - Propagates a `_reversion_user_id` (if provided) into task message headers for downstream use.
+    - Wraps task execution in a Reversion revision, attaching the user (when resolvable)
+      and a structured JSON comment describing the task call.
+    """
+
+    abstract = True
+
+    def apply_async(self, args: Any | None = None, kwargs: Any | None = None, **options) -> Any:
```
Struchid: `None` is a part of `Any` already: 

```suggestion
    def apply_async(self, args: Any | None = None, kwargs: Any = None, **options) -> Any:
```

vitrina/middleware.py
```diff
@@ -23,3 +27,35 @@ def __call__(self, request):
...
+        self, request: HttpRequest, view_func: Any, view_args: list[Any], view_kwargs: dict[str, Any]
+    ) -> None:
+        if not reversion.is_active():
+            return None
+
+        if reversion.get_comment():
+            return None
+
+        if hasattr(view_func, "model_admin"):
+            return None
```
Struchid: These can be combined with `or` since they do the same thing

vitrina/utils.py
```diff
@@ -39,3 +48,76 @@ def translate_text(text: str, field_name: str = "") -> str | None:
...
+
+def is_model_versioned(model: type[models.Model]) -> bool:
+    """Return whether the given Django model should be treated as versioned.
+
+    A model is considered versioned if:
+    - it belongs to an installed app whose dotted name starts with "vitrina" (via `get_all_models()`), and
+    - its fully-qualified name (<module>.<class>) does not match any pattern in
+      `settings.NOT_VERSIONED_MODELS`.
+    """
+    if model not in get_all_models(app_prefix="vitrina"):
```
Struchid, line 62: Models should not change that often, we could use a cache for that. Redis maybe (with a TTL) ? Or have we already discussed this place, I do not remember which one it was

vitrina/resources/models.py
gts446: Good point! Updated.

vitrina/utils.py
gts446: It was different part. Good point adding cache'ing here. Updated.

vitrina/admin.py
```diff
@@ -0,0 +1,140 @@
...
+    extra = 0
+    can_delete = False
+
+    fields = (
+        "version_data",
+        "column_content_type",
+        "column_object_id",
+        "column_object_repr",
+    )
+    readonly_fields = fields
```
martynasracct:
```suggestion
    fields = readonly_fields = (
        "version_data",
        "column_content_type",
        "column_object_id",
        "column_object_repr",
    )
    readonly_fields = fields
```

vitrina/apps.py
```diff
@@ -1,6 +1,40 @@
...
+
+    def _register_models_for_reversion(self, versioned_models: set[type[models.Model]]) -> None:
+        try:
+            for model in versioned_models:
+                if not reversion.is_registered(model):
+                    reversion.register(model)
+
+        except (OperationalError, ProgrammingError) as error:
+            logger.warning(f"Error during django-reversion model registration: {error}")
+            pass
```
martynasracct: `pass` can be deleted

vitrina/celery.py
```diff
@@ -1,10 +1,73 @@
...
+
+    def apply_async(self, args: Any = None, kwargs: Any = None, **options) -> Any:
+        """Enqueue the task, optionally embedding `_reversion_user_id` into message headers."""
+        kwargs = kwargs or {}
+
+        user_id = kwargs.pop("_reversion_user_id", None)
+
+        headers = options.pop("headers", {})
+        if user_id:
+            headers = {**headers, "_reversion_user_id": user_id}
```
martynasracct:
```suggestion
        headers = options.pop("headers", {})
        if user_id := kwargs.pop("_reversion_user_id", None)
            headers = {**headers, "_reversion_user_id": user_id}
```

vitrina/celery.py
```diff
@@ -1,10 +1,73 @@
...
+
+        return super().apply_async(args=args, kwargs=kwargs, headers=headers, **options)
+
+    def __call__(self, *args, **kwargs) -> Any:
+        """Execute the task inside a Reversion revision.
+
+        Tries to read `u_reversion_user_id` from Celery request headers, resolves it to a Django user,
```
martynasracct:
```suggestion
        Tries to read `_reversion_user_id` from Celery request headers, resolves it to a Django user,
```

vitrina/celery.py
```diff
@@ -1,10 +1,73 @@
...
+        and sets Reversion's user accordingly. Always writes a JSON comment describing
+        the invocation (task name + args/kwargs).
+        """
+        with reversion.create_revision():
+            request = getattr(self, "request", None)
+            user_id = None
+            if request:
+                headers = getattr(request, "headers", {}) or {}
```
martynasracct: `user_id = None`  is not needed, since it is reassigned in line 52
```suggestion
            if request := getattr(self, "request", None)
                headers = getattr(request, "headers", {}) or {}
```

vitrina/celery.py
```diff
@@ -1,10 +1,73 @@
...
+
+            if user_id := headers.get("_reversion_user_id"):
+                User = get_user_model()
+                try:
+                    user = User.objects.get(pk=user_id)
+                except User.DoesNotExist:
+                    user = None
+
+                if user:
+                    reversion.set_user(user)
```
martynasracct, line 53:
```suggestion
                try:
                    user = User.objects.get(pk=user_id)
                    reversion.set_user(user)
                except User.DoesNotExist:
                    pass
```

martynasracct (approved):
Translations are missing.

Also, I suggest getting at least 2-3 reviews on this PR before merging and test changes thoroughly, since the solution you chose to implement, in my opinion, is not very django-ish.

vitrina/admin.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def has_change_permission(self, request: HttpRequest, obj: Revision | None = None) -> bool:
+        return False
+
+    def has_delete_permission(self, request: HttpRequest, obj: Revision | None = None) -> bool:
+        return False
+
+    @admin.display(description=_("Versijos duomenys"))
+    def version_data(self, obj: Version) -> str:
+        data = json.loads(obj.serialized_data)[0]["fields"]
```
wisecrow: It is good practice to handle possible errors: `JSONDecodeError`, `IndexError`, or `KeyError`.

vitrina/admin.py
```diff
@@ -0,0 +1,139 @@
...
+from django.utils.html import format_html
+import reversion
+from reversion.models import Revision, Version
+from reversion.admin import VersionAdmin
+
+from vitrina.utils import RevisionComment, RevisionSource
+from django.utils.translation import gettext_lazy as _
+
+
+class VersionInline(admin.TabularInline):
```
wisecrow, line 21: There could be hundreds rows loaded here and sometimes it freezes when too much rows are trying to load.
Consider adding https://docs.djangoproject.com/en/6.0/ref/contrib/admin/#django.contrib.admin.InlineModelAdmin.max_num

vitrina/apps.py
```diff
@@ -1,6 +1,39 @@
...
 
 
 class ApiConfig(AppConfig):
     name = "vitrina"
     label = "vitrina"
+
+    def ready(self) -> None:
+        super().ready()
+        eligable_models = get_all_models(app_prefix="vitrina")
+        self._validate_not_versioned_patterns(eligable_models)
```
wisecrow: Validation runs on every app load. On every migration, management or worker command. It can break migrations. I suggest to restrict it only to dev environment:
```
if settings.DEBUG:
        self._validate_not_versioned_patterns(eligable_models)
```

vitrina/apps.py
```diff
@@ -1,6 +1,39 @@
...
+logger = logging.getLogger(__name__)
 
 
 class ApiConfig(AppConfig):
     name = "vitrina"
     label = "vitrina"
+
+    def ready(self) -> None:
+        super().ready()
+        eligable_models = get_all_models(app_prefix="vitrina")
```
wisecrow:
```suggestion
        eligible_models = get_all_models(app_prefix="vitrina")
```

vitrina/apps.py
```diff
@@ -1,6 +1,39 @@
...
+        versioned_models = {model for model in eligable_models if is_model_versioned(model)}
+        self._register_models_for_reversion(versioned_models)
+
+    def _register_models_for_reversion(self, versioned_models: set[type[models.Model]]) -> None:
+        try:
+            for model in versioned_models:
+                if not reversion.is_registered(model):
+                    reversion.register(model)
+
+        except (OperationalError, ProgrammingError) as error:
```
wisecrow: Do we really want to fail silently? It could be serious problem we want to fix before proceeding further.

vitrina/utils.py
```diff
@@ -39,3 +49,77 @@ def translate_text(text: str, field_name: str = "") -> str | None:
...
+
+    @classmethod
+    def from_json(cls, raw_json: str) -> "RevisionComment | None":
+        try:
+            data = json.loads(raw_json)
+        except json.JSONDecodeError:
+            return None
+
+        if "source" not in data:
+            raise ValueError("Missing 'source' in revision comment")
```
wisecrow, line 115: Inconsitent error handling: returns error in failure and raises `ValueError`. Both cases means `corrupt` data and we should treat it as one problem and use single way to handle it: return `None` or raise `ValueError` but not both.

vitrina/middleware.py
```diff
@@ -23,3 +27,29 @@ def __call__(self, request):
...
+    and the view arguments, if a revision is active.
+    """
+
+    def process_view(
+        self, request: HttpRequest, view_func: Any, view_args: list[Any], view_kwargs: dict[str, Any]
+    ) -> None:
+        if not reversion.is_active() or reversion.get_comment() or hasattr(view_func, "model_admin"):
+            return None
+
+        match = request.resolver_match
```
wisecrow, line 44: Can be `None` in case of `404` error or maybe other error cases.
```suggestion
        match = request.resolver_match
        if not match:
            return None
```

vitrina/celery.py
```diff
@@ -1,10 +1,68 @@
...
+        """Execute the task inside a Reversion revision.
+
+        Tries to read `_reversion_user_id` from Celery request headers, resolves it to a Django user,
+        and sets Reversion's user accordingly. Always writes a JSON comment describing
+        the invocation (task name + args/kwargs).
+        """
+        with reversion.create_revision():
+            request = getattr(self, "request", None)
+            if request:
+                headers = getattr(request, "headers", {}) or {}
```
wisecrow, line 47: Undefined `headers`.  Will crash outside `if request` scope.
Add `headers = {}` in the beginning of the function.

wisecrow (changes requested):
Some issues needs to be fixed

vitrina/admin.py
gts446: Number of rows depends on how many models have been modified in single "action/revision". So in theory it shouldn't be that many. Usually 1-2, in some occasions several.
Regarding your suggested method, it does not limit number of displayed inlines that already exist, it only affects adding new rows, and works a bit weird. Lets say currently there is 7 rows, and I have set `get_max_num` to 10, so it prerenders 3 empty rows and forbids to add new :).
I think it is safe here not to limit number of inlines(another option could be to use 3rd party library for pagination), what do you say?

vitrina/apps.py
gts446: I removed this try/catch completely, it was supposed to only raise those exceptions during migrations, but it does not, so no need to handle them here.

vitrina/utils.py
gts446: Good point, updated.

vitrina/celery.py
gts446: Good catch. Updated.

vitrina/admin.py
wisecrow: Let's leave as is now and later we will see if any limitation is needed.

wisecrow (approved):

PR #2192, opened 2025-12-09
Ft/versionable tables duplicates fields

### LINKED ISSUE #2194: Versijavimas PART2 + PART3.
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/2190
#

Tėvinė užduotis:
- https://github.com/atviriduomenys/katalogas/issues/1997

### Rolė, tikslas, funkcionalumas

```markdown
Kaip naudotojas,
tam, kad galėčiau versijuoti savo duomenų struktūrą,
noriu, kad turėčiau pasirinkimą atsidaryti norimą versiją ir publikuoti teisingos struktūros versiją.
```

### Aprašymas

Versijavimui reikalingi puslapiai, kurie rodytų tik tuos objektus, kurie yra konkrečioje versijoje.
Versijavimui reikalingas publikavimo pakeitimas, kad publikuojama versija būtų sukuriama ir išsaugoma pilnai.

### As-Is (jei taikoma)

Visi struktūros puslapiai labai ribotai palaiko versijų rodymą. Rodomos visos versijos vienu metu, sunku diferencijuoti tarp versijų.

Publikavimas pilnų versijų nekuria, saugo tik skirtumus, neįmanoma arba labai sunku keisti jau sukurtą struktūrą

### To-Be (pagrindinis)

Užduotys:
- [x] Kiekvienas puslapis (kuriam versijos aktualu) palaiko `version_id` parametrą URL'e. [Part 2]
- [x] Versijos pasirinkimo mygtukas su drop-down versijų menu. [Part 2]
- [x] Breadcrumbs atnaujinti, kad aiškiai matytųsi kurios versijos puslapyje esame. [Part 2]
- [x] Versijos publikavimas sukuria pilną versiją ir ją saugo Metadata lentelėje ir visose su ja susijusiose lentelėse. Versijos nepriklauso viena nuo kitos [Part 3]

Publikavimo pavyzdinis validavimas:
- [ ] Metadata duomenų rinkinys (Dataset) privalo būti publikuojamas.
- [ ] Property referencina modelį, kuris yra jo datasete, tačiau kitoje versijoje ir yra Draft. Property publikavimas neleidžiamas su error message.
- [ ] Property referencina modelį, kuris yra jo datasete, toje pačioje versijoje ir yra Draft. Property publikavimas leidžiamas, jei Useris publikuoja ir referencinamą modelį. Jei ne -- error message.
- [ ] Property referencina modelį, kuris yra jo dataset, kitoje versijoje, tačiau yra Publikuotas. Property publikavimas leidžiamas.
- [ ] Property referencina modelį, kuris yra kitame datasete, bet yra Draft versijoje. Property publikavimas neleidžiamas
- [ ] Property referencina modelį, kuris yra kitam datasete, bet yra Publikuotas. Property publikavimas leidžiamas.

Gustascct:
Most of the diff comes from rewritten tests

This PR introduces logic for duplicating fields when creating new versions. 

Logic:
- Each Metadata instance is connected to one of 7 (Base, Model, Property, EnumItem, ParamItem, Dataset, DatasetDistribution) tables through object + content_type (I'll call these tables Related table). 
- Each Metadata instance must be duplicated. 
- Each related table instance must be duplicated.
- Every needed foreign key relationship from a related table to another table must be handled by either creating a duplicate of that instance or by setting a new ID.

Implementation:
- Each Metadata instance that was selected to publish is duplicated. Its related table instance is also duplicated.
- The related table instances are stored in a dictionary, old_instance: new_instance
- Then we take a new related instance and iterate through its foreign key relationships. If the relationship is needed we handle it like this:
- - First we validate the field. 
- - If its valid: 
- - - We check if its EnumItem or ParamItem. If it is then we go to Enum or Param table, duplicate the field and set the primary keys. Then using object + content_type (Enum and Param have these) we check if the related table is Dataset. If it is we ignore it. If not, using our dictionary, set a new object id.
- - - If its not, we set the new value of the foreign key relationship by using our dictionary. If the field passes validation then it must be in that dictionary.
- - If its not valid we return an error message (error messages will probably change).


tests/structure/test_views.py
Gustascct, line 1: Can skip, most of the diff is from rewriting old tests. If you want to review, the new tests start at 4984

vitrina/structure/migrations/0012_version_number_assigning.py
Gustascct, line 1: Previous migration created version instances with no version number, this fixes it.

vitrina/structure/migrations/0012_version_number_assigning.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 4.2.23 on 2025-12-08 08:27
+
+from django.db import migrations
+
+
+def assign_version_number_if_missing(apps, schema_editor):
+    Version = apps.get_model("vitrina_structure", "Version")
+
+    datasets_with_bad_versions = Version.objects.filter(version__isnull=True).values_list("dataset_id", flat=True)
```
Struchid: Nitpick:


```suggestion
    datasets_with_no_versions = Version.objects.filter(version__isnull=True).values_list("dataset_id", flat=True)
```

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+                    )
+                except ValidationError as e:
+                    transaction.set_rollback(True)
+                    form.add_error(None, e.message)
+                    return self.form_invalid(form)
+
+        if self.new_version:
+            return redirect(reverse("dataset-structure", args=[self.dataset.pk, self.new_version.pk]))
+        else:
+            return redirect(reverse("dataset-structure", args=[self.dataset.pk, self.metadata_version.pk]))
```
Struchid: I'd set the second argument in a variable like:

`version_pk = self.new_version.pk if self.new_version else self.metadata_version.pk`

And use that as an argument:


```suggestion
        version_pk = self.new_version.pk if self.new_version else self.metadata_version.pk
        return redirect(reverse("dataset-structure", args=[self.dataset.pk, version_pk]))
```

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+
+    def create_related_model_duplicate(self, old_metadata_instance):
```
Struchid: Typehints here and in the other methods

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
-                    base=meta.object.base if isinstance(meta.object, Model) else None,
-                    status=meta.status if meta.status else None,
+        selected_metadata = form.cleaned_data.get("metadata", [])
+        old_to_new_metadata_object_map = {}
+        with transaction.atomic():
+            self.new_version.save()
+
+            for meta in selected_metadata:
+                if meta := Metadata.objects.filter(pk=meta).first():
+                    # Duplicate metadata row
```
Struchid: Comments should answer the question "Why" and not "What", it can be seen from the method that you will be copying the metadata row `create_metadata_duplicate` but the reason for "Why" is not seen from this.

vitrina/structure/views.py
```diff
@@ -3363,6 +3372,9 @@ def get_context_data(self, **kwargs):
         return context
 
     def form_valid(self, form):
```
Struchid: Not your but might as well add typehints since you're touching the function and make it better for future generations :-)

vitrina/structure/views.py
```diff
@@ -3363,6 +3372,9 @@ def get_context_data(self, **kwargs):
...
+        if self.metadata_version.status != VersionStatus.DRAFT:
+            form.add_error(None, "Publikuota versija negali būti publikuota dar kartą.")
```
Struchid: Shouldn't this be translatable too?


```suggestion
            form.add_error(None, _("Publikuota versija negali būti publikuota dar kartą."))
```

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+                    )
+
+            already_created_fields = old_to_new_metadata_object_map
+            for old_related_instance, new_related_instance in list(old_to_new_metadata_object_map.items()):
+                try:
+                    already_created_fields = self.duplicate_foreign_key_relationships(
+                        new_related_instance, already_created_fields
+                    )
+                except ValidationError as e:
+                    transaction.set_rollback(True)
```
Struchid, line 3634: So the emails will be sent but half of the code will not be executed, is that OK? If not, why not move the email sending below (if the changes are not related to the content of the email, ignore this)

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+            return redirect(reverse("dataset-structure", args=[self.dataset.pk, self.new_version.pk]))
+        else:
+            return redirect(reverse("dataset-structure", args=[self.dataset.pk, self.metadata_version.pk]))
+
+    def create_related_model_duplicate(self, old_metadata_instance):
+        related_model = old_metadata_instance.content_type.model_class()
+        if related_model == Dataset:
+            return None
+        old_related_instance = related_model.objects.filter(id=old_metadata_instance.object_id).first()
+        new_related_instance = deepcopy(old_related_instance)
```
Struchid, line 3645: I would double-check if deepcopy is really the intended/suggested by django way of copying an object, something tells me it's not (but might just be me:D)

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+            EnumItem,
+            ParamItem,
+            DatasetDistribution,
+            Enum,
+            Param,
+            Prefix,
+        ]
+        for field in new_related_object._meta.get_fields():
+            if isinstance(field, ForeignKey):
+                # If the foreign key relationship is not in the needed list skip
```
Struchid: That's quite understandable from the code, no real value from this message, unless you think it improves readability, then ignore.

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+
+                if isinstance(deeper_old_related_object, (Enum, Param)):
+                    deeper_new_related_object = deepcopy(deeper_old_related_object)
+                    deeper_new_related_object.pk = None
+                    deeper_new_related_object.metadata_version = self.new_version
+
+                    if hasattr(deeper_new_related_object, "content_type_id"):
+                        related_model = deeper_new_related_object.content_type.model_class()
+                        if related_model in [Property, Model, DatasetDistribution]:
+                            if deeper_new_related_object.object not in already_created_fields:
```
Struchid: These two can be combined

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+        new_metadata_instance.draft = False
+        if old_metadata_instance.status == Status.objects.filter(codename=StatusCode.DEVELOP).first():
+            new_metadata_instance.status = Status.objects.filter(codename=StatusCode.COMPLETED).first()
+
+        new_metadata_instance.metadata_version = self.new_version
+        new_metadata_instance.save()
+
+        return old_metadata_instance, new_metadata_instance
+
+    def duplicate_foreign_key_relationships(self, new_related_object, already_created_fields):
```
Struchid: A lot is happening in this one, if possible - make a couple helpers for this to separate it out a bit (and to avoid nesting as much as possible)

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+            return None
+        old_related_instance = related_model.objects.filter(id=old_metadata_instance.object_id).first()
+        new_related_instance = deepcopy(old_related_instance)
+        new_related_instance.pk = None
+        new_related_instance.metadata_version = self.new_version
+        new_related_instance.save()
+
+        return old_related_instance, new_related_instance
+
+    def create_metadata_duplicate(self, old_metadata_instance) -> tuple:
```
martynasracct: type hints missing

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+        new_metadata_instance = deepcopy(old_metadata_instance)
+        new_metadata_instance.pk = None
+        # TODO: column draft can be deprecated after full versioning is completed.
+        new_metadata_instance.draft = False
+        if old_metadata_instance.status == Status.objects.filter(codename=StatusCode.DEVELOP).first():
```
martynasracct: I don't really like `Status.objects.filter(codename=StatusCode.DEVELOP).first()` query. It leaves a lot of assumptions:
- Why `.first()`? Can there be more than one `Status` object with `StatusCode.DEVELOP`? 
- If not - maybe this should be `Status.objects.get(codename=StatusCode.DEVELOP)` or just a value saved in `self`?
- If yes - why we only compare status to first one
- It also assumes that if `old_metadata_instance` doesn't have status and `Status` does not exist - it's also fine. Is this really the case?

Maybe it would be better to write it this way? (But I am not sure if this is correct for this use case)
```suggestion
        if old_metadata_instance.status and old_metadata_instace.status.codename == StatusCode.DEVELOP:
```

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+            new_metadata_instance.status = Status.objects.filter(codename=StatusCode.COMPLETED).first()
```
martynasracct: Similar thing here: why `.first()`? Is it fine that value is `None` if `Status` does not exist, etc. 

Also, it's called inside a loop, so it adds 1 query for every loop iteration. Maybe better solution:
```python
@cached_property
def get_status_completed() -> Status:
    return get_object_or_404(Status, codename=StatusCode.COMPLETED)

...

new_metadata_instance.status = get_status_completed()
```

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+        if self.new_version:
+            return redirect(reverse("dataset-structure", args=[self.dataset.pk, self.new_version.pk]))
+        else:
+            return redirect(reverse("dataset-structure", args=[self.dataset.pk, self.metadata_version.pk]))
+
+    def create_related_model_duplicate(self, old_metadata_instance):
+        related_model = old_metadata_instance.content_type.model_class()
+        if related_model == Dataset:
+            return None
+        old_related_instance = related_model.objects.filter(id=old_metadata_instance.object_id).first()
```
martynasracct: Is there a reason why `GenericForeignKey` cannot be used here?

```suggestion
        if type(old_metadata_instance.object) == Dataset:
            return None
        old_related_instance = old_metadata_instance.object
```

vitrina/structure/views.py
```diff
@@ -3424,34 +3435,212 @@ def form_valid(self, form):
...
+                if meta := Metadata.objects.filter(pk=meta).first():
+                    # Duplicate metadata row
+                    old_metadata_instance, new_metadata_instance = self.create_metadata_duplicate(meta)
+
+                    # Duplicate related object
+                    related_object_duplication_result = self.create_related_model_duplicate(old_metadata_instance)
+                    if related_object_duplication_result:
+                        old_related_instance, new_related_instance = related_object_duplication_result
+                    else:
+                        continue
```
martynasracct:
```suggestion
                    if not related_object_duplication_result := self.create_related_model_duplicate(old_metadata_instance):
                        continue
                    old_related_instance, new_related_instance = related_object_duplication_result

```

vitrina/structure/migrations/0012_version_number_assigning.py
Gustascct: Changed

vitrina/structure/views.py
Gustascct: Nice, changed

vitrina/structure/views.py
Gustascct: added

vitrina/structure/views.py
Gustascct: Removed most comments, they were explaining code that needed no explanation

vitrina/structure/views.py
Gustascct: added

vitrina/structure/views.py
Gustascct: Yes, now it is, thanks

vitrina/structure/views.py
Gustascct: I found this answer on stackoverflow which references a discussion about this, so just used it. seems to work fine

https://stackoverflow.com/questions/8917983/how-can-i-create-a-deep-clone-of-a-db-object-in-django

vitrina/structure/views.py
Gustascct: Combined

vitrina/structure/views.py
Gustascct: added

vitrina/structure/views.py
Gustascct: yeah good point, the statuses are created in a migration so they should always exist.

vitrina/structure/views.py
Gustascct: implemented, thanks

vitrina/structure/views.py
Gustascct: Did not know that was possible, but isinstance should be used instead of type. Implemented

vitrina/structure/views.py
Gustascct: Implemented, thanks

vitrina/structure/views.py
Gustascct: Tried to split it as much as possible and to remove nesting without making it worse
If its still barely readable, I will try something else

vitrina/structure/views.py
Gustascct: Not sure what the emails are about, will ask in another meeting and will create a fix PR if thats okay with you

wisecrow:
There are few bugs i found while testing this branch in my local dev env.:
1. Pre release "version" is decreased when draft "version" ins increased after publication:
Screenshot 2025-12-30 at 10 38 54
2. In structure page, after picking "draft" version - non models is listed, while pre release version has model listed in model structure view.

vitrina/structure/views.py
wisecrow: if the outer transaction isn't properly managed, already-saved objects might not be rolled back.
Suggestion: wrap the entire duplication logic in an explicit transaction.atomic() block at the method level:
```
@transaction.atomic
def form_valid(self, form):
```

vitrina/structure/views.py
```diff
@@ -3542,34 +3557,238 @@ def form_valid(self, form):
...
+    def check_if_field_has_same_dataset(self, field: RELATED_OBJECT_TYPE) -> bool:
+        return field.metadata_version.dataset == self.dataset
+
+    def check_if_field_has_same_version(self, field: RELATED_OBJECT_TYPE) -> bool:
+        return field.metadata_version == self.metadata_version
+
+    def check_if_field_has_published_version(self, field: RELATED_OBJECT_TYPE) -> bool:
+        return field.metadata_version.status and field.metadata_version.status != VersionStatus.DRAFT
+
+    def check_if_field_is_valid(
```
wisecrow, line 3744: This method is ambiguous. The method signature suggests it returns `None`, but it's used for side effects (raising exceptions).  
Either rename to `validate_field_relationships()` or return a boolean and let the caller handle the exception

vitrina/structure/views.py
```diff
@@ -3542,34 +3557,238 @@ def form_valid(self, form):
...
+                            new_related_object.metadata.first().prepare or new_related_object.metadata.first()
+                        )
+                        error_msg = _(
+                            "Laukas {0} turi nuorodą į nepublikuojamą lauką tame pačiame duomenų ištekliuje.".format(
+                                error_field_name
+                            )
+                        )
+                        raise ValidationError(error_msg)
+
+                    if type(deeper_new_related_object.object) is not Dataset:
```
wisecrow, line 3707: Using `type()` is generally discouraged as it doesn't respect inheritance.

```suggestion
                   if isinstance(deeper_old_related_object, (Enum, Param)):
```

wisecrow (approved):

PR #2209, opened 2025-12-12
Fix celery on test env

### LINKED ISSUE #2151: Celery neveikia Test aplinkoj
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/2198
#

## Priešistorė
Su https://github.com/atviriduomenys/katalogas/pull/2013 buvo pridedas background procesu sprendimas kad leisti `spinta check` kaip fonine užduotį, nes gali užtrukti ilgiau, nei vartotojas gali laukti.

## Problema
Lokaliai demonstruojant viskas veikia, tačiau sukėlus i test aplinką (https://test.data.gov.lt) gaunama klaida bandant sukurti nauja foninę `spinta check` užduotį. 

### Klaidos atkūrimas
Nuėjus https://test.data.gov.lt/admin/vitrina_structure/manifestvalidationentry/ ir bandant prideti nauja įraša su DSA failu kurį reikia patikrinti:

Image

Saugant gaunama klaida:

Image

Iš klaidos galima pamatyti jog nepavyksta prisijungti RabbitMQ (kaip message broker):

Image

Kas iškart identifikuoja bėda jog jungiami į neteisinga broker, nes naudojamas Redis (valkey), o ne RabbitMq, bei adresas `amqp://guest:**@127.0.0.1:5672//` yra neteisingas.

### Užduoties rezultatas
Surasti CELERY konfiguracijos klaidas,
jei jos Python kodo bazėje ir gali būti sukontroliuotos per repositorijos pakeitimus - juos sutaisyti.
jei jos del infrastruktūros (env variables, dns pasiekiamumas ir pan) - sukurti užduoti Laurynui.

gts446:
Fixes issue where Django used default Celery broker settings and tried to connect to the wrong broker.
Also updated docker settings, because apparently Celery was failing to run before.
Also added a fix to create task only after changes are commited to database, to avoid error, where Celery tries to execute task before Manifest actually exists.

### DIFF
--- a/vitrina/structure/admin.py
+++ b/vitrina/structure/admin.py
@@ -1,6 +1,7 @@
 from django.contrib import admin
 from django.forms import ModelForm
 from django.http import HttpRequest
+from django.db import transaction
 
 from vitrina.structure.models import Prefix, ManifestValidationEntry
 from vitrina.structure.tasks import validate_manifest_task
@@ -16,7 +17,7 @@ class ManifestValidationEntryAdmin(admin.ModelAdmin):
 
     def save_model(self, request: HttpRequest, obj: ManifestValidationEntry, form: ModelForm, change: bool) -> None:
         super().save_model(request, obj, form, change)
-        validate_manifest_task.delay(obj.uuid)
+        transaction.on_commit(lambda: validate_manifest_task.delay(obj.uuid))
 
 
 admin.site.register(Prefix, PrefixAdmin)
--- a/docker-compose.template.yml
+++ b/docker-compose.template.yml
@@ -69,6 +69,9 @@ services:
       - katalogas
 
   katalogas-${BRANCH}-celery:
+    build:
+      context: .
+      dockerfile: docker/Dockerfile
     image: katalogas-${BRANCH}:latest
     command: celery -A vitrina worker -l info
     environment:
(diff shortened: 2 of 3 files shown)


wisecrow (approved):

vitrina/structure/admin.py
```diff
@@ -16,7 +17,7 @@ class ManifestValidationEntryAdmin(admin.ModelAdmin):
 
     def save_model(self, request: HttpRequest, obj: ManifestValidationEntry, form: ModelForm, change: bool) -> None:
         super().save_model(request, obj, form, change)
-        validate_manifest_task.delay(obj.uuid)
+        transaction.on_commit(lambda: validate_manifest_task.delay(obj.uuid))
```
Struchid, line 20: Can this be test-proven? E.g. to show that the task is not called if the transaction is roll-backed & the task is called after a successful commit

Struchid (approved):

vitrina/structure/admin.py
gts446: I guess it can. This way would be to mitigate atomic transaction which saves the model and then fails, and test if the function was called (as you described). But wouldn't this mean that test is testing if "django works as it should"?
Another approach would be to to test django admin "view" and see if task was initiated after submitting form.
This wouldn't test my change explicitly, but I guess would fix missing test problem for this functionality.
We could also add integration test, which actually tests if task was sent to broker, then Celery picked it, and executed it. But not sure if that "beast" should/can be tested with unit tests...
What do you think?

Struchid: Well it is nice to prove everything with tests if we're going by "what should be done", but in this case I guess we can discuss this thing with the team and merge for now. And then write tests later if everybody agrees.

PR #2229, opened 2025-12-23
Prod deployment fixes and improvements

### LINKED ISSUE #2224: Portalo nesklandumai po naujos versijos išleidimo
Portalo klaidos po naujos kataalogo versijos išleidimo.

**Gamybinė:**
- [x] https://github.com/atviriduomenys/katalogas/issues/2230
- [x] https://github.com/atviriduomenys/katalogas/issues/2255
- [ ] #2256
- [x] https://github.com/atviriduomenys/katalogas/issues/2248
- [x] https://github.com/atviriduomenys/katalogas/issues/2238
- [x] https://github.com/atviriduomenys/spinta/issues/1588
- [x] #2257
- [x] https://github.com/atviriduomenys/katalogas/issues/2261
- [x] Neišeina sukurti rinkinio, meta 500 klaidą Sutaisyta
- [x] Super userio aplinkoje portalo viršuje rodomas admin aplinkos meniu sąrašas
- [x]  VDA taip pat pranešė, kad negali pridėti struktūros aprašo, meta 504 klaidą Sutaisyta
- [x]  502 klaida bandant atidaryti portalą, redaguoti rinkinį, pridėti ADSA ir kt.
- [x] Pasikartojanti 500 klaida `/datasets` endpointe
- [x] Neranda `django-cms` ir `fotnawesome` css failų prode. Sutaisyta.

**Testinė:**
- [x] 500 klaida bandant atidaryti rinkinių sąrašą (laikinas sprendimas `django-admin update_index`) https://github.com/atviriduomenys/katalogas/discussions/1907

### LINKED ISSUE #2227: Titulinio langelių suvienodinimas
Reikia sulyginti šitas korteles, kad gražiai atrodytų.

Image

Dabar jos prisitaiko prie teksto apimties, jeigu daug prirašyta, tai langelis bus didelis, jeigu mažai - mažas.

Taip padaryta turbūt dėl to, kad yra rinkinių su labai ilgais pavadinimais, todėl tiesiog jeigu default būtų padarytas mažas langelis, pavadinimas dingtų už langelio ribų, o dabar prisitaiko prie pavadinimo ilgio, bet atrodo nelabai gražiai. Reikia padaryti vienodą ir maksimalų langelio dydį.

kastger:
- Make django logging config configurable from `.env`.
- Allow organization updates for superusers.
- Fix cards on landing page.
- Add custom error pages.
- Revert `CustomSignalProcessor` (needs followup task to improve performance and get rid of N+1 problem, maybe even move to celery, better to revert the changes to make sure everything is atleast working as it was before).
- Clear crontabs.

image

image


.env.example
```diff
@@ -11,6 +11,7 @@ SPINTA_SERVER_URL=https://127.0.0.1:8000
 SPINTA_SERVER_NAME=127.0.0.1:8000
 SPINTA_PATH=spinta-env/data
 USE_OTP_VALIDATION=True
+DJANGO_LOGGING={"version": 1, "disable_existing_loggers": false, "handlers": {"console": {"level": "DEBUG", "class": "logging.StreamHandler"}}, "root": {"handlers": ["console"], "level": "WARNING"}}
```
martynasracct, line 14: I would suggest separate settings file for test and for prod, with different `LOGGING` configurations, instead of setting them via env file. 

This way it's easier to maintain it. And all developers can check what is logged for test / prod

karina-klinkeviciute (approved):
operatyviai padirbėta, ačiū 👍🏻

vitrina/orgs/views.py
```diff
@@ -132,7 +132,7 @@ def get_context_data(self, **kwargs):
         context_data = super().get_context_data(**kwargs)
         context_data["can_view_members"] = has_perm(self.request.user, Action.VIEW, Representative, self.organization)
         context_data["can_view_contacts"] = has_perm(self.request.user, Action.VIEW, Contact, self.organization)
-        context_data["can_update_organization"] = (
+        context_data["can_update_organization"] = self.request.user.is_superuser or (
             has_perm(self.request.user, Action.UPDATE, Representative, self.organization)
             and self.request.user.viisp_organization == self.organization
         )
```
martynasracct: `has_perm()` already returns `True` if user is superuser.

```suggestion
        context_data["can_update_organization"] = (
            has_perm(self.request.user, Action.UPDATE, Representative, self.organization)
            and (self.request.user.viisp_organization == self.organization or self.request.user.is_superuser)
        )
```

vitrina/locale/en/LC_MESSAGES/django.po
martynasracct, line 1: Please use following commands to generate translation strings
```
poetry run python ../manage.py makemessages -l en --no-location
poetry run python ../manage.py makemessages -l lt --no-location
```

vitrina/locale/lt/LC_MESSAGES/django.po
martynasracct, line 1: Please use following commands to generate translation strings
```
poetry run python ../manage.py makemessages -l en --no-location
poetry run python ../manage.py makemessages -l lt --no-location
```

vitrina/settings.py
```diff
@@ -453,20 +453,24 @@
...
 ACCOUNT_SIGNUP_REDIRECT_URL = "password-set"
 
-LOGGING = {
-    "version": 1,
-    "disable_existing_loggers": False,
-    "handlers": {
-        "console": {
-            "level": "INFO",
-            "class": "logging.StreamHandler",
+LOGGING = env.json(
```
martynasracct, line 456: Same suggestion about separate settings files

vitrina/status_codes_views.py
martynasracct, line 1: I suggest naming this file with `_views.py` suffix, So it's in line with other django view files

martynasracct (approved):

vitrina/orgs/views.py
kastger: True that `has_perm()` already checks superuser, but my version is better because it short-circuits - if the user is a superuser, we get True immediately without calling has_perm() at all. Your version calls has_perm() unnecessarily, which then checks is_superuser again internally. 

Why call a function when we already know the answer?

.env.example
kastger: Since our deployment is still not automated I wanted the logging config to be more flexible so that we don't need to redeploy after config change. We currently have a problem from `spinta` that it overrides the root config and logs INFO and DEBUG levels. As per your suggestion we would need to merge new configs to `devel` branch and redeploy. I agree that we should be able to always be able to check for test / prod configurations, but without automated deployment this is too long when production is getting flooded.

karina-klinkeviciute: ChatGPT recommends having one settings file with only levels in env variables, not the full logging structure

kastger: What about when I need to add new handler? For example I want separate options for elasticsearch, different from root. That would require changes to `settings.py` and redeployment.

martynasracct: Redeployment shouldn't be an issue. If it is (and it usually is in Katalogas) it's a different problem that (in my opinion) has to be solved.

And about having only debug levels in env variables, I see the same problem, that parts of logging config are still hidden from developer. But there is no reason to hide it.

kastger: Without setting logging config or level in env variables it can be difficult to investigate a problem - what if I need to check different level logs in test/prod? My point is log level changes shouldn't require redeployment (waiting CI/CD pipelines, packaging and after investigation - rollback the config, make the fix and deploy again)? Config for test/prod shouldnt be different anyway and it would always be kept as default in `settings.py` so everyone knows what is used, but in an emergency like this last deployment it can be changed, help prevent a service crash and give us more time for a proper fix.

vitrina/orgs/views.py
lescijuscct: "Why call a function when we already know the answer?"
To avoid duplicating business logic.

lescijuscct:
This PR is quite large and combines multiple unrelated changes.
It would be much easier to review, reason about, and test if the work was split into smaller, focused PRs, each addressing a single concern.

vitrina/orgs/views.py
lescijuscct: What is the purpose of this change?  The commit message states that it changes logic (Let update organization for super users. ) ,  Martynas's comment raises concern that this change is redundant. Is there a reason why you want to keep it?

kastger: The purpose of this change is to allow superusers to modify organizations. The current code returns false for `viisp_organization == self.organization`, but superusers shouldn't care about `viisp_organization`.

martynasracct: Change is not redundant. For super-user: `has_perm(...)` -> `True` and `viisp_organization` check -> `False`, . So change @kastger made actually makes this check to pass for super users.

But I still think it is a "code smell" at least.

kastger:
> This PR is quite large and combines multiple unrelated changes. It would be much easier to review, reason about, and test if the work was split into smaller, focused PRs, each addressing a single concern.

Agreed,I cramped everything into one because production is fragile and I wanted to push fixes faster as they are pretty small (except for the reverted `CustomSignalProcessor`), but they are unrelated to each other.

vitrina/orgs/views.py
kastger: "To avoid duplicating business logic"  this might be true, but I think this was originally overlooked and put in as mistake - why are superusers only allowed to modify `viisp_organization` when they should be allowed to modify all? This change makes superuser > permission system.

lescijuscct: Since the superuser will now bypass the viisp_organization check, I don’t see a strong reason for keeping this logic outside of has_perm.

I’d suggest moving the viisp_organization check into has_perm itself, rather than spreading this logic across multiple places. This would centralize permission handling and make the behavior easier to reason about and maintain.

lescijuscct: In addition, after this change, there is an inconsistency in behavior.

In OrganizationUpdateView.has_permission:

```
def has_permission(self):
    org = self.get_object()
    return has_perm(self.request.user, Action.UPDATE, org) and self.request.user.viisp_organization == org

```

Even though the superuser bypasses the viisp_organization check inside has_perm, this explicit comparison means the superuser will not bypass the viisp_organization check here.

This is another reason why the viisp_organization logic should live entirely inside has_perm, so permission behavior is consistent and not accidentally overridden at the view level.

kastger: Agreed with code smell. I will try to see if its possible to move `viisp_organization` check to `has_perm`, that way we would get rid of the smell and keep the permission check centralized.

kastger: @martynasracct @lescijuscct Made changes to `has_perm` added tests.

tests/orgs/test_services.py
```diff
@@ -1191,3 +1191,201 @@ def test_permission_with_organization_representative_for_all_related_datasets(se
...
+            organization=organization,
+            content_type=ct,
+            object_id=organization.pk,
+            user=None,
+            role=Representative.COORDINATOR
+        )
+
+        # This mimics what OrganizationUpdateView.has_permission() does.
+        res = has_perm(user, Action.UPDATE, organization)
+        assert res is True
```
lescijuscct: is this test a duplicate for test_coordinator_with_matching_viisp_org_can_update?

tests/orgs/test_services.py
```diff
@@ -1191,3 +1191,201 @@ def test_permission_with_organization_representative_for_all_related_datasets(se
...
+            content_type=ct,
+            object_id=organization.pk,
+            role=Representative.COORDINATOR
+        )
+        coordinator.user.is_viisp_login = True
+        coordinator.user.viisp_company_code = other_org.company_code
+        coordinator.user.save()
+
+        res = has_perm(coordinator.user, Action.UPDATE, organization)
+        assert res is True
```
lescijuscct, line 1292: I’m not fully confident about this, but my understanding is that this test may not align with the business rules described by the PO. Unfortunately, those rules are not formally written down, which makes it hard to be certain.

vitrina/orgs/services.py
```diff
@@ -610,14 +610,23 @@ def has_perm(
...
                 user_org
                 and Representative.objects.filter(
                     where,
                     organization=user_org,
                 ).exists()
             ):
+                if model == Organization and action == Action.UPDATE:
+                    user_viisp_org = getattr(user, "viisp_organization", None)
+                    if isinstance(obj, Type):
+                        return True
```
lescijuscct: Could you help me understand this logic?

This code is under the Organization + UPDATE branch. 

1) In which cases would obj be a type rather than an instance here, and why would we want to bypass the check during an update?
1.1) Can obj legitimately be a type (not an instance) for the UPDATE action?

2) Why does this logic exist only for UPDATE?
2.1) Why does CREATE bypass this viisp check entirely?

.env.example
karina-klinkeviciute: Maybe redeployment shouldn't be an issue, but it is now. But even if we had an actual continuous deployment, it might be an issue - we would need to wait for someone to approve a PR if we need to change logging level.

vitrina/orgs/views.py
karina-klinkeviciute: Martynas didn't raise concern that this code is redundant, he suggested to move the check to a different place. 

Original:
```
context_data["can_update_organization"] = self.request.user.is_superuser or (
            has_perm(self.request.user, Action.UPDATE, Representative, self.organization)
            and self.request.user.viisp_organization == self.organization
        )
```

Geraldas's code:

```
context_data["can_update_organization"] = self.request.user.is_superuser or (
            has_perm(self.request.user, Action.UPDATE, Representative, self.organization)
            and self.request.user.viisp_organization == self.organization
        )
```

Martynas's suggestion:

```
context_data["can_update_organization"] = (
            has_perm(self.request.user, Action.UPDATE, Representative, self.organization)
            and (self.request.user.viisp_organization == self.organization or self.request.user.is_superuser)
```

i.e. the addition `or self.request.user.is_superuser` remains, only Martynas suggested to move it a different place. 

Geraldas's point is early exit, and Martynas's point is not duplicating business logic. Both are valid. In this case, though, I'd roughly evaluate the cost of not exiting early, and if it's not huge (i.e. not some heavy queries or things like that), I would go with Martynas's sugestion, because not duplicating business logic is important for avoiding mistakes and for readability.

tests/orgs/test_services.py
karina-klinkeviciute: As I understand, this is the behaviour that is needed for VDA for open data? I'm not sure how we want to continue with this. I think, every organisation will need to assign their own coordinator anyway. So in this case, I think, we should lose this option for someone to change the data for another organisation. 

But yes, I don't think it's written anywhere. We need to confirm this with the PO. 

I think we need some document on rights management, documenting who can do what, which would cover all cases. We have a couple of tasks, documenting representatives' rights on resources, maybe the information in them (the rights matrices, role descriptions, exceptions) could be extended to those cases also?

tests/orgs/test_services.py
kastger: My bad, It was duplicate.

vitrina/orgs/services.py
kastger: I removed the part that checks `isinstance`, you are right - since the code is for UPDATE action.

It exists only for UPDATE because the problematic area in this PR was `viisp_organization` check. I moved the check part into `has_perm` and made it apply only for the same situation it was used. CREATE bypassed the viisp check entirely before aswell, because the viisp check wasn't used anywhere else (or maybe I missed it).

tests/orgs/test_services.py
kastger: This test passes with the old code too, I was just trying to keep the same behavior  in tact.

karina-klinkeviciute: I think this should be left like this for now, and later we will discuss it with Mantas

kastger:
Splt into separate PRs:
https://github.com/atviriduomenys/katalogas/pull/2254
https://github.com/atviriduomenys/katalogas/pull/2253
https://github.com/atviriduomenys/katalogas/pull/2252
https://github.com/atviriduomenys/katalogas/pull/2251
https://github.com/atviriduomenys/katalogas/pull/2250

PR #2233, opened 2025-12-30
Ft/versionable tables resource fix

### LINKED ISSUE #2243: Versioning changes for DatasetDistribution
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/2306
#

### Poreikis

```markdown
Kaip naudotojas,
noriu galėti versijuoti duomenų distribucijas ir aiškiai matyti bendras distribucijas ir konkrečios versijos distribucijas.
```

### Siekiama būsena

Duomenų distribucijų rodymo lange turi matytis versijos dropdown menu, kuriame galima būtų pasirinkti versijas.
Duomenų distribucijoms turi būti suteikiama versija tik kai ta pati duomenų distribucija yra struktūros apraše.
Duomenų distribucijos kurios nėra struktūros apraše turi būti bendros ir matytis visose versijose.

### Wireframes

Pridėtas versijų pasirinkimo dropdown menu. Pasirinkus versiją matomos tos versijos duomenų distribucijos, bei visos bendros duomenų distribucijos
Image

Matoma, kad versijai pasikeitus duomenų distribucija Saugykla liko, bet modeliai kuriems Saugykla kuria duomenų distribucijas priklauso konkrečiai versijai.
Image

### Acceptance Criteria

- [x] Duomenų ištekliaus lange, prie duomenų distribucijų atvaizdavimo formos turi būti pridėtas versijų dropdown menu leidžiantis pasirinkti versiją.
- [x] Duomenų distribucijų redagavimo ir atvaizdavimo formos turi palaikyti ir versijuotas duomenų distribucijas ir ne versijuotas duomenų distribucijas.
- [x] Duomenų distribucija tampa versijuojama tik tada, kai ji yra referencinama iš modelio.
- [x] Duomenų distribucija, kuri nėra versijuota turi matytis visose versijose, tačiau. neturi būti įtraukta į struktūros aprašą (Metadata lentelę)
- [x] Modelių redagavimo ir kūrimo languose turi atsirasti galimybė pasirinkti duomenų distribuciją, kuri nėra Saugyklos API, kuri dar neturi versijos arba turi tokią pačią versiją kaip modelis.
- [x] Duomenų distribucijos atvaizdavimo lange turi būti pašalintas Naujas Modelis mygtukas, jeigu naujo modelio kurti neįmanoma dėl versijos nebuvimo.

### 📋 Definition of Done

- [ ] **Testai** parašyti / atnaujinti
- [ ] **Dokumentacija atnaujinta** (jei reikia)
- [x] Padeployinta ir ištestuota **Laikinojoj testinėj** aplinkoj
- [x] Acceptance Criteria įvykdyti
- [ ] CHANGES.rst
- [x] Pademonstruota per Sprint Review
- [ ] PO (ar atsakingas asmuo) peržiūrėjo ir patvirtino

Gustascct:
This PR does:

1. Fixes a bug where importing DSA with no resource automatically created a Saugyklos API distribution and assigned it to models.
2. Implements versioning for resource pages and resources themselves. This is done in this manner:
- - Creating a DatasetDistribution **does not** assign it a metadata_version and does not create a metadata instance.
- - When a DatasetDistribution is referenced by a model, then it **gets** the metadata_version and its metadata instance.
- - When all models stop referencing a DatasetDistribution, its metadata instance gets **deleted** and its metadata_version gets **removed**.
3. Creates two migrations:
- 1. First one creates a version for all datasets that do not have one yet.
- 2. The second one creates a Dataset metadata instance for datasets which do not have one yet. It also fixes DatasetDistributions, by following the rules outlined in 1 and 2.
4. Updates the DatasetDistribution display window. Adds a version picker which lets you choose the version and shows the DatasetDistributions which have the same version **or** the DatasetDistributions that do not have a version.

**Workaround for Metadata name:** The name generating function does not enforce the limit of 255, so to make the migration work I have increased the limit to 400. I checked out the Datasets which were hitting this limit on the test environment, I could not produce the error that I can produce it on my data dump.

Tests coming in another PR


vitrina/datasets/services.py
```diff
@@ -432,9 +435,9 @@ def _generate_download_url(self, models, distribution_format):
 
         return download_url, distribution_name
 
-    def retrieve_data(self, dataset_pk, resource_name, distribution_format):
+    def retrieve_data(self, dataset_pk, metadata_version, resource_name, distribution_format):
```
martynasracct: typing-hints missing

vitrina/datasets/templates/vitrina/datasets/detail.html
```diff
@@ -158,11 +158,14 @@
         <div class="columns no-margin-bottom" id="resources">
             <div class="column is-full">
                 <h2 class="custom-title is-size-4-mobile no-margin-bottom">{% translate "Duomenų šaltiniai" %}</h2>
-                {% include 'vitrina/datasets/datatable.html' %}
+                {% include 'vitrina/datasets/datatable.html'%}
```
martynasracct:
```suggestion
                {% include 'vitrina/datasets/datatable.html' %}
```

vitrina/datasets/views.py
```diff
@@ -459,6 +459,21 @@ def has_permission(self):
         dataset = get_object_or_404(Dataset, id=self.kwargs["pk"])
         return has_perm(self.request.user, Action.VIEW, dataset)
 
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        metadata_version = request.GET.get("resource_version")
+        if not metadata_version:
+            metadata_version = _Version.objects.filter(dataset=self.object).order_by("version").last()
+
+            if metadata_version:
```
martynasracct:
```suggestion
            if metadata_version := _Version.objects.filter(dataset=self.object).order_by("version").last()
```

vitrina/datasets/views.py
```diff
@@ -459,6 +459,21 @@ def has_permission(self):
         dataset = get_object_or_404(Dataset, id=self.kwargs["pk"])
         return has_perm(self.request.user, Action.VIEW, dataset)
 
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        metadata_version = request.GET.get("resource_version")
+        if not metadata_version:
```
martynasracct:
```suggestion
        if not metadata_version := request.GET.get("resource_version")
```

vitrina/datasets/views.py
```diff
@@ -459,6 +459,21 @@ def has_permission(self):
...
+            metadata_version = _Version.objects.filter(dataset=self.object).order_by("version").last()
+
+            if metadata_version:
+                url = (
+                    reverse("dataset-detail", kwargs={"pk": self.object.pk})
+                    + f"?resource_version={metadata_version.pk}"
+                )
```
martynasracct, line 484:
```suggestion
                url = f'{reverse("dataset-detail", kwargs={"pk": self.object.pk})}?resource_version={metadata_version.pk}'
```

vitrina/datasets/views.py
```diff
@@ -470,6 +485,13 @@ def get_queryset(self) -> QuerySet[Dataset]:
     def get_context_data(self, **kwargs):
         context_data = super().get_context_data(**kwargs)
         dataset = context_data.get("dataset")
+        dataset_versions = _Version.objects.filter(dataset=dataset).order_by("version")
+        context_data["versions"] = dataset_versions
```
martynasracct:
```suggestion
        context_data["versions"] = _Version.objects.filter(dataset=dataset).order_by("version")
```

vitrina/datasets/views.py
```diff
@@ -470,6 +485,13 @@ def get_queryset(self) -> QuerySet[Dataset]:
...
+        metadata_version_id = self.request.GET.get("resource_version")
+        metadata_version = None
+        if metadata_version_id:
```
martynasracct:
```suggestion
        metadata_version = None
        if metadata_version_id := self.request.GET.get("resource_version")
```

vitrina/resources/models.py
```diff
@@ -129,7 +130,7 @@ class DatasetDistribution(TranslatableModel):
     version = models.IntegerField(default=1)
     deleted = models.BooleanField(blank=True, null=True)
     deleted_on = models.DateTimeField(blank=True, null=True)
-    dataset = models.ForeignKey(Dataset, models.CASCADE)
+    dataset = models.ForeignKey("vitrina_datasets.Dataset", models.CASCADE)
```
martynasracct, line 136: This change was needed because of circular import?

vitrina/resources/models.py
```diff
@@ -342,7 +354,18 @@ def get_acl_parents(self):
         return parents
 
     def get_absolute_url(self):
```
martynasracct:
```suggestion
    def get_absolute_url(self) -> str:
```

vitrina/resources/models.py
```diff
@@ -400,3 +423,56 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+                name = (
+                    Metadata.objects.filter(
+                        dataset=self.dataset,
+                        content_type=ContentType.objects.get_for_model(DatasetDistribution),
+                        name__iregex=r"resource[0-9]+",
+                    )
+                    .order_by("name")
+                    .values_list("name", flat=True)
+                    .last()
+                )
```
martynasracct: `name__iregex=r"resource[0-9]+",` looks scary. Why only `Metadata` objects with these symbols specifically? What other `Metadata` names can be there?

vitrina/resources/models.py
```diff
@@ -400,3 +423,56 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+
+    def create_metadata_instance_and_assign_version(self, metadata_version: int) -> Metadata:
+        metadata_version = Version.objects.get(pk=metadata_version)
+        if metadata_instance := self.metadata.first():
+            metadata_instance.metadata_version = metadata_version
+            metadata_instance.save()
+        else:
+            name = self.name
+            if not name:
+                name = (
```
martynasracct:
```suggestion
                latest_metadata_name = (
```

vitrina/resources/models.py
```diff
@@ -400,3 +423,56 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+                if not name:
+                    name = "resource1"
+                else:
+                    n = name.replace("resource", "")
+                    try:
+                        n = int(n)
+                    except ValueError:
+                        n = 0
+                    n += 1
+                    name = f"resource{n}"
```
martynasracct: This part is confusing, I suggest moving it into separate function. Function should make this code more descriptive. For example:

```python
def _get_next_metadata_name(previous_metadata_name: str | None) -> str:
    """
    Get version from previous_metadata_name and increase it by 1. Return 1 if previous_metadata_name is None
    """
    if not previous_metadata_name:
        previous_metadata_name = "resource0"

    version_string = previous_metadata_name.replace("resource", "")
    version_number = int(version_string) if version_string.isdigit() else 0
    
    return f"resource{version_number + 1}"
```

It's just an example, maybe you will think of a better way to do this

vitrina/resources/urls.py
```diff
@@ -11,20 +11,40 @@
...
     path("resource/<int:pk>/add", ResourceCreateView.as_view(), name="resource-add"),
-    path("resource/<int:pk>/change", ResourceUpdateView.as_view(), name="resource-change"),
-    path("resource/<int:pk>/delete", ResourceDeleteView.as_view(), name="resource-delete"),
+    path("resource/<int:pk>/versions/<int:version_id>/change", ResourceUpdateView.as_view(), name="resource-change"),
     path(
-        "datasets/<int:pk>/resource/<int:resource_id>",
+        "resource/<int:pk>/change",
+        ResourceUpdateView.as_view(),
+        name="resource-change-no-version",
+    ),
```
martynasracct: This URL should fit into single line

vitrina/resources/views.py
```diff
@@ -104,42 +102,6 @@ def form_valid(self, form):
         resource.dataset = self.dataset
         resource.save()
```
martynasracct, line 112: Just checking: you moved this code into `create_metadata_instance_and_assign_version` method. Don't you need to call this method here?

vitrina/resources/views.py
```diff
@@ -343,10 +306,11 @@ def has_permission(self):
         dataset = get_object_or_404(Dataset, id=self.kwargs["pk"])
         return has_perm(self.request.user, Action.VIEW, dataset)
 
-    def get_data(self, dataset_pk, model_name, distribution_format):
+    def get_data(self, dataset_pk, metadata_version_pk, model_name, distribution_format):
```
martynasracct: type checking missing

vitrina/structure/migrations/0012_populate_version_of_new_tables.py
martynasracct, line 1: Are these migrations already merged to main branch?

vitrina/structure/migrations/0013_version_number_assigning.py
martynasracct, line 1: Are these migrations already merged to main branch?

vitrina/structure/models.py
```diff
@@ -304,14 +303,45 @@ def get_base_props(self):
...
 
+    def save(self, *args, **kwargs):
+        old_distribution = None
+
+        if self.pk:
+            old_instance = Model.objects.get(pk=self.pk)
+            old_distribution = old_instance.distribution
+
+        if self.distribution:
+            if self.distribution.format.extension == "UAPI":
```
martynasracct:
```suggestion
        if self.distribution and self.distribution.format.extension == "UAPI":
```

vitrina/resources/migrations/0012_datasetdistribution_access_datasetdistribution_level_and_more.py
```diff
@@ -0,0 +1,44 @@
...
+
+def populate_new_distribution_fields(apps, schema_editor):
+    DatasetDistribution = apps.get_model('vitrina_resources', 'DatasetDistribution')
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    Metadata = apps.get_model('vitrina_structure', 'Metadata')
+    dd_ct = ContentType.objects.get_for_model(DatasetDistribution)
+
+    for dataset_distribution in DatasetDistribution.objects.all():
+        metadata_instance = Metadata.objects.filter(content_type=dd_ct, object_id=dataset_distribution.id).first()
+        if metadata_instance:
```
martynasracct:
```suggestion
        if metadata_instance := Metadata.objects.filter(content_type=dd_ct, object_id=dataset_distribution.id).first():
```

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
```diff
@@ -0,0 +1,29 @@
...
+
+from vitrina.structure import VersionStatus
+
+
+def create_version_for_datasets_with_no_version(apps, schema_editor):
+    Dataset = apps.get_model('vitrina_datasets', 'Dataset')
+    Version = apps.get_model('vitrina_structure', 'Version')
+
+    for dataset in Dataset.objects.all():
+        if not dataset.dataset_version.exists():
```
martynasracct:
```suggestion
    for dataset in Dataset.objects.filter(dataset_version__isnull=True).prefetch_related("metadata_set"):

```

vitrina/structure/migrations/0013_dataset_datasetdistribution_data_fix.py
martynasracct, line 1: Scary migration 😨

martynasracct (approved):

vitrina/datasets/services.py
Gustascct: Added

vitrina/datasets/templates/vitrina/datasets/detail.html
Gustascct: changed

vitrina/datasets/views.py
Gustascct: changed

vitrina/datasets/views.py
Gustascct: changed

vitrina/datasets/views.py
Gustascct: changed

vitrina/resources/models.py
Gustascct: Yes it was

vitrina/resources/models.py
Gustascct: changed

vitrina/resources/models.py
Gustascct: I took this entire code from another place, so it was not written by me. Should I rewrite it?

vitrina/resources/models.py
Gustascct: changed

vitrina/resources/views.py
Gustascct: changed

vitrina/resources/views.py
Gustascct: No, the logic changed. Now a resource is created unversioned and without a metadata row. Only after it is referenced the version and metadata get created

vitrina/structure/migrations/0012_populate_version_of_new_tables.py
Gustascct: Not merged on devel

vitrina/structure/migrations/0013_version_number_assigning.py
Gustascct: Not merged on devel

vitrina/structure/models.py
Gustascct: Changed

vitrina/resources/migrations/0012_datasetdistribution_access_datasetdistribution_level_and_more.py
Gustascct: Changed

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
Gustascct: Changed

vitrina/datasets/views.py
Gustascct: Changed

vitrina/datasets/views.py
Gustascct: Changed

vitrina/resources/urls.py
Gustascct: Changed

vitrina/datasets/templates/vitrina/datasets/detail.html
```diff
@@ -162,7 +162,10 @@ <h2 class="custom-title is-size-4-mobile no-margin-bottom">{% translate "Duomen
             </div>
         </div>
     {% else %}
-        {% include 'vitrina/datasets/add_resource.html' %}
+        <div class="buttons is-right m-t-md">
+            {% include "version_picker.html" with version_as_query_param=True url_name="dataset-detail" dataset=dataset versions=versions selected_version=selected_version %}
+            {% include 'vitrina/datasets/add_resource.html' %}
```
Struchid: Let's keep consistent


```suggestion
            {% include "vitrina/datasets/add_resource.html" %}
```

vitrina/resources/migrations/0012_datasetdistribution_access_datasetdistribution_level_and_more.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def populate_new_distribution_fields(apps, schema_editor):
+    DatasetDistribution = apps.get_model('vitrina_resources', 'DatasetDistribution')
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    Metadata = apps.get_model('vitrina_structure', 'Metadata')
+    dd_ct = ContentType.objects.get_for_model(DatasetDistribution)
+
+    for dataset_distribution in DatasetDistribution.objects.all():
+        if metadata_instance := Metadata.objects.filter(content_type=dd_ct, object_id=dataset_distribution.id).first():
```
Struchid: This is the n+1 problem, even though these are migrations, might be worth improving on it

vitrina/resources/models.py
```diff
@@ -341,8 +353,19 @@ def get_acl_parents(self):
             parents.extend(self.dataset.get_acl_parents())
         return parents
 
-    def get_absolute_url(self):
-        return reverse("resource-detail", kwargs={"pk": self.dataset.pk, "resource_id": self.pk})
+    def get_absolute_url(self) -> str:
+        if self.metadata_version and self.metadata_version.pk:
```
Struchid, line 359: Is there a possibility that `metadata_version` will be unsaved for the `DatasetDistribution`, or is this (`self.metadata_version.pk`) more of a sanity check?

vitrina/resources/models.py
```diff
@@ -400,3 +423,56 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
 
         for entry in all_entries.filter(url__in=new_urls):
             entry.update_description()
+
+    def create_metadata_instance_and_assign_version(self, metadata_version: int) -> Metadata:
+        metadata_version = Version.objects.get(pk=metadata_version)
```
Struchid: Let's add some safeguards here, if the Version you would try to `get()` does not exist, otherwise this will fail hard for the user.

Let's add logs for that too

vitrina/resources/models.py
```diff
@@ -400,3 +423,56 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+                        name__iregex=r"resource[0-9]+",
+                    )
+                    .order_by("name")
+                    .values_list("name", flat=True)
+                    .last()
+                )
+                if not latest_metadata_name:
+                    name = "resource1"
+                else:
+                    n = latest_metadata_name.replace("resource", "")
```
Struchid: Let's not use one letter variables (if we're already touching that part of the code), e.g.:

```suggestion
                    duplicate_name_suffix = latest_metadata_name.replace("resource", "")
```

vitrina/resources/models.py
```diff
@@ -400,3 +423,56 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+                    except ValueError:
+                        n = 0
+                    n += 1
+                    name = f"resource{n}"
+                self.name = name
+            self.save()
+            metadata_instance = Metadata.objects.create(
+                uuid=str(uuid.uuid4()),
+                dataset=self.dataset,
+                content_type=ContentType.objects.get_for_model(DatasetDistribution),
```
Struchid, line 442: Called twice, could be called once at the beginning of the method

vitrina/resources/models.py
```diff
@@ -400,3 +423,56 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+                else:
+                    n = latest_metadata_name.replace("resource", "")
+                    try:
+                        n = int(n)
+                    except ValueError:
+                        n = 0
+                    n += 1
+                    name = f"resource{n}"
+                self.name = name
+            self.save()
```
Struchid: Seems like `update_fields` could be used, due to only updating one attribute

vitrina/resources/models.py
```diff
@@ -400,3 +423,56 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
 
         for entry in all_entries.filter(url__in=new_urls):
             entry.update_description()
+
+    def create_metadata_instance_and_assign_version(self, metadata_version: int) -> Metadata:
```
Struchid: The method seems to not necessarily create new metadata instance, but might re-use an older instance

vitrina/resources/views.py
```diff
@@ -370,9 +335,13 @@ def get(self, request, *args, **kwargs):
...
-            "structure_url": reverse("dataset-structure", args=[self.dataset.pk]),
-            "data_url": reverse("model-data", args=[self.dataset.pk, self.models[0].name]) if self.models else None,
-            "api_url": reverse("getall-api", args=[self.dataset.pk, self.models[0].name]) if self.models else None,
+            "structure_url": reverse("dataset-structure", args=[self.dataset.pk, metadata_version_pk]),
+            "data_url": reverse("model-data", args=[self.dataset.pk, metadata_version_pk, self.models[0].name])
+            if self.models
+            else None,
+            "api_url": reverse("getall-api", args=[self.dataset.pk, metadata_version_pk, self.models[0].name])
+            if self.models
+            else None,
```
Struchid: I'd extract the `self.models[0].name` above to avoid duplication

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
```diff
@@ -0,0 +1,28 @@
...
+def create_version_for_datasets_with_no_version(apps, schema_editor):
+    Dataset = apps.get_model('vitrina_datasets', 'Dataset')
+    Version = apps.get_model('vitrina_structure', 'Version')
+
+    for dataset in Dataset.objects.filter(dataset_version__isnull=True).prefetch_related("metadata_set"):
+            created_draft_version = Version.objects.create(dataset=dataset, version=1, status=VersionStatus.DRAFT)
+            created_draft_version.save()
+            for metadata in dataset.metadata_set.all():
+                metadata.metadata_version = created_draft_version
+                metadata.save()
```
Struchid: Seems like this is indented 2x, unless it is github with its formatting nonsenses again

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
```diff
@@ -0,0 +1,28 @@
...
+from vitrina.structure import VersionStatus
+
+
+def create_version_for_datasets_with_no_version(apps, schema_editor):
+    Dataset = apps.get_model('vitrina_datasets', 'Dataset')
+    Version = apps.get_model('vitrina_structure', 'Version')
+
+    for dataset in Dataset.objects.filter(dataset_version__isnull=True).prefetch_related("metadata_set"):
+            created_draft_version = Version.objects.create(dataset=dataset, version=1, status=VersionStatus.DRAFT)
+            created_draft_version.save()
```
Struchid: You are already doing create and not just instantiating an instance of the Version model, no need to do the save once more.

vitrina/structure/forms.py
```diff
@@ -463,6 +464,12 @@ class ModelCreateForm(forms.ModelForm):
         widget=forms.RadioSelect,
         help_text=_("Savybė nurodanti modelio metaduomenų gyvavimo ciklo būseną."),
     )
+    distribution = forms.ModelChoiceField(
+        label=_("Duomenų distribucija"),
```
Struchid, line 468: IMO It should translate to `pateiktis` instead of `distribucija`, but I might be missing some context and therefore lying

vitrina/structure/models.py
```diff
@@ -304,14 +303,44 @@ def get_base_props(self):
...
-            for type in Metadata.ACCESS_TYPES:
+            for type in AccessType.choices:
                 if type[0] == access:
                     return type[1]
         return ""
 
     def is_opened(self):
         return self.dataset.is_opened()
 
+    def save(self, *args, **kwargs):
```
Struchid: Let's add return typehints for the methods in this file

vitrina/structure/views.py
```diff
@@ -3647,6 +3654,30 @@ def create_related_model_duplicate(self, old_metadata_instance: Metadata) -> tup
         new_related_instance.metadata_version = self.new_version
         new_related_instance.save()
 
+        if isinstance(old_related_instance, DatasetDistribution):
+            for trans in old_related_instance.translations.all():
```
Struchid: Let's use full names instead, wouldn't want to cause any confusion what we're looping through 😶

```suggestion
            for translation in old_related_instance.translations.all():
```

Struchid (approved):

wisecrow:
Kai bandau importuoti naują struktūrą https://github.com/atviriduomenys/demo-saltiniai/blob/main/manifest/datasets/gov/vssa/demo/wsdl/wsdl_rctest.csv
gaunu klaidą- 
```
File "/app/vitrina/structure/services.py", line 155, in _load_datasets
_link_distributions(meta, dataset, metadata_version)
File "/app/vitrina/structure/services.py", line 636, in _link_distributions
model.save()
File "/app/vitrina/structure/models.py", line 321, in save
if self.distribution and self.distribution.format.extension == "UAPI":
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'extension'
```

vitrina/datasets/services.py
```diff
@@ -321,19 +322,22 @@ class DynamicResourceService:
...
 
     def generate_resources(self, is_for_rdf_export=False):
         if not self.dataset.is_part_of_dataservice():
             return []
 
-        models = self.dataset.model_set.all()
+        if is_for_rdf_export:
+            models = self.dataset.model_set.filter()
+        else:
+            models = self.dataset.model_set.filter(metadata_version=self.metadata_version)
```
gts446:
```suggestion
        models = self.dataset.model_set.filter()
        if is_for_rdf_export:
            models = models.filter(metadata_version=self.metadata_version)
```

vitrina/datasets/views.py
```diff
@@ -470,6 +482,11 @@ def get_queryset(self) -> QuerySet[Dataset]:
     def get_context_data(self, **kwargs):
         context_data = super().get_context_data(**kwargs)
         dataset = context_data.get("dataset")
+        context_data["versions"] = _Version.objects.filter(dataset=dataset).order_by("version")
+        metadata_version = None
+        if metadata_version_id := self.request.GET.get("resource_version"):
```
gts446: Is it possible that this will be None?

vitrina/datasets/views.py
```diff
@@ -459,6 +459,18 @@ def has_permission(self):
...
 
+    def get(self, request, *args, **kwargs):
+        self.object = self.get_object()
+
+        if not request.GET.get("resource_version"):
+            if metadata_version := _Version.objects.filter(dataset=self.object).order_by("version").last():
+                url = (
+                    f"{reverse('dataset-detail', kwargs={'pk': self.object.pk})}?resource_version={metadata_version.pk}"
+                )
+                return redirect(url)
```
gts446: Might wanna add `else` here to validate that version actually exists here, otherwise user can pass anything here and later parts will fail.

gts446, line 485:
```suggestion
        if resource_version := request.GET.get("resource_version"):
            try:
                self.metadata_version = _Version.objects.get(pk=resource_version)
                return super().get(request, *args, **kwargs)
            except:
                pass
       
        if metadata_version := _Version.objects.filter(dataset=self.object).order_by("version").last():
            url = (
                f"{reverse('dataset-detail', kwargs={'pk': self.object.pk})}?resource_version={metadata_version.pk}"
            )
            return redirect(url)
        self.metadata_version = None
```

Also you can remove later code in this view that retrieves `metadata_version` and use `self.metadata_version`

vitrina/datasets/views.py
```diff
@@ -470,6 +482,11 @@ def get_queryset(self) -> QuerySet[Dataset]:
     def get_context_data(self, **kwargs):
         context_data = super().get_context_data(**kwargs)
         dataset = context_data.get("dataset")
+        context_data["versions"] = _Version.objects.filter(dataset=dataset).order_by("version")
+        metadata_version = None
+        if metadata_version_id := self.request.GET.get("resource_version"):
+            metadata_version = _Version.objects.filter(id=metadata_version_id).first()
+            context_data["selected_version"] = metadata_version
```
gts446, line 504:
```suggestion
        context_data["selected_version"] = self.metadata_version
```

vitrina/datasets/views.py
```diff
@@ -511,8 +527,19 @@ def get_context_data(self, **kwargs):
             "licences": set([dist.licence for dist in dataset.datasetdistribution_set.filter(licence__isnull=False)]),
         }
 
+        if metadata_version.status == VersionStatus.DRAFT:
```
gts446: This will crash if metadata_version is None.

Also update instances of `metadata_version` to `self.metadata_version`

vitrina/resources/migrations/0012_datasetdistribution_access_datasetdistribution_level_and_more.py
```diff
@@ -0,0 +1,43 @@
...
+            dataset_distribution.name = metadata_instance.name
+            dataset_distribution.access = metadata_instance.access
+            dataset_distribution.level = metadata_instance.level
+            dataset_distribution.save()
```
gts446: As I understand here `Metadata` can be linked with `DatasetDistribution` through `GenericForeighKey`? In that case it would make sense to add new virtual field `metadata = GeneriRelation('Metadata')` to `DatasetDistribution`. and then change code to 
```suggestion
    for dataset_distribution in DatasetDistribution.objects.filter(metadata__isnull=False).select_related("metadata"):
        if metadata_instance := dataset_distribution.metadata.first():
            dataset_distribution.name = metadata_instance.name
            dataset_distribution.access = metadata_instance.access
            dataset_distribution.level = metadata_instance.level
            dataset_distribution.save()
```

vitrina/resources/models.py
```diff
@@ -400,3 +423,32 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+        metadata_version = Version.objects.get(pk=metadata_version)
+        if metadata_instance := self.metadata.first():
```
gts446:
```suggestion
    def create_metadata_instance_and_assign_version(self, metadata_version_id: int) -> Metadata:
        metadata_version = Version.objects.get(pk=metadata_version_id)
        if metadata_instance := self.metadata.first():
```

vitrina/resources/models.py
```diff
@@ -400,3 +423,32 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+            metadata_instance.metadata_version = metadata_version
+            metadata_instance.save()
```
gts446:
```suggestion
            metadata_instance.save(update_fields=["metadata_version"])
```

vitrina/resources/models.py
```diff
@@ -400,3 +423,32 @@ def update_applicable_legislation(self, urls: list[str]) -> None:
...
+                metadata_version=metadata_version,
+            )
+
+        return metadata_instance
+
+    def check_if_resource_should_be_versioned(self):
+        if not self.model_set.exists():
+            self.metadata_version = None
+            self.save()
+            self.metadata.all().delete()
```
gts446: By the name of this method it seems that this checks something and returns True/False or raises excepton.
My suggestion would be to do this check `if not self.model_set.exists():` whereever you need this method. and then maybe rename and change this method to:

```suggestion
    def delete_metadata(self):
        self.metadata_version = None
        self.save()
        self.metadata.all().delete()
```

Or leave as is but only rename method to `delete_metadata_if_has_no_models`.

vitrina/resources/views.py
```diff
@@ -280,16 +231,22 @@ def handle_no_permission(self):
             return redirect(settings.LOGIN_URL)
         else:
             resource = get_object_or_404(DatasetDistribution, id=self.kwargs["pk"])
+            if resource.metadata_version:
+                url = reverse("dataset-detail", kwargs={"pk": resource.dataset.pk})
+                return HttpResponseRedirect(f"{url}?resource_version={resource.metadata_version.pk}")
```
gts446: I suggest to move `resource` from variable to instance property inside `has_permission` and then call it wherever needed. Becuase now it is fetched at least three times (in `has_permisison`, `handle_no_permission`, `form_valid`)

vitrina/resources/views.py
```diff
@@ -280,16 +231,22 @@ def handle_no_permission(self):
             return redirect(settings.LOGIN_URL)
         else:
```
gts446, line 267: no need for `else` here

vitrina/resources/views.py
```diff
@@ -307,6 +264,11 @@ def form_valid(self, form: BaseForm) -> HttpResponse:
                     user=self.request.user,
                 )
                 dataset.save()
+
+        if version_id:
+            url = reverse("dataset-detail", kwargs={"pk": dataset.pk})
+            return HttpResponseRedirect(f"{url}?resource_version={version_id}")
```
gts446:
```suggestion
        with create_revision():
            add_to_revision(resource)
            set_user(self.request.user)
            set_comment((f'Ištrintas duomenų šaltinis "{resource.lt_title()}".'))
            resource.delete()
            resource.metadata.all().delete()

            if not DatasetDistribution.objects.filter(dataset=dataset) and dataset.is_public:
                if dataset.plandataset_set.exists():
                    dataset.status = Dataset.PLANNED
                    comment_status = Comment.PLANNED
                else:
                    dataset.status = Dataset.INVENTORED
                    comment_status = Comment.INVENTORED

                Comment.objects.create(
                    content_type=ContentType.objects.get_for_model(dataset),
                    object_id=dataset.pk,
                    type=Comment.STATUS,
                    status=comment_status,
                    user=self.request.user,
                )
                dataset.save()

        if  version_id := self.kwargs.get("version_id"):
            url = reverse("dataset-detail", kwargs={"pk": dataset.pk})
            return HttpResponseRedirect(f"{url}?resource_version={version_id}")
```

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
```diff
@@ -0,0 +1,28 @@
...
+from django.db import migrations
+
+from vitrina.structure import VersionStatus
+
+
+def create_version_for_datasets_with_no_version(apps, schema_editor):
+    Dataset = apps.get_model('vitrina_datasets', 'Dataset')
+    Version = apps.get_model('vitrina_structure', 'Version')
+
+    for dataset in Dataset.objects.filter(dataset_version__isnull=True).prefetch_related("metadata_set"):
```
gts446, line 34: Isnt `Dataset.version` deprecated? Shouldn't you check by `dataset__metadata__metadata_version`?

vitrina/structure/models.py
```diff
@@ -304,14 +303,42 @@ def get_base_props(self):
...
+        old_distribution = None
+
+        if self.pk:
+            old_instance = Model.objects.get(pk=self.pk)
+            old_distribution = old_instance.distribution
+
+        if self.distribution and self.distribution.format and self.distribution.format.extension == "UAPI":
+            raise ValidationError(_("Negalima priskirti Saugyklos API distribucijos. Pasirinkite kitą distribuciją."))
```
gts446: This should be at the top of the method.

vitrina/structure/models.py
```diff
@@ -304,14 +303,42 @@ def get_base_props(self):
...
+
+        super().save(*args, **kwargs)
+
+        if self.distribution:
+            if self.distribution.metadata_version and self.distribution.metadata_version != self.metadata_version:
+                return
```
gts446, line 325: `DatasetDistribution` should not have duplicated `metadata_version` field which is directly linked with `DatasetDistribution.metadata.metadata_version`

vitrina/structure/models.py
```diff
@@ -304,14 +303,42 @@ def get_base_props(self):
...
+            self.distribution.create_metadata_instance_and_assign_version(self.metadata_version.pk)
+            self.distribution.metadata_version = self.metadata_version
+            self.distribution.save(update_fields=["metadata_version"])
+
+        if old_distribution:
+            old_distribution.check_if_resource_should_be_versioned()
```
gts446: Looking at this whole method I hardly understand what are you trying to achieve here.
I can see you validate that Model is not assigned to Saugykla, but that should be handled in form validations not in model.save().
Also could you explain the rest of the logic?
Why you create metadata for distribution only when it is assigned to model?
If that is some business requirement then I think this should also be done inside view, not model.save()

gts446 (changes requested):

vitrina/datasets/services.py
Struchid: The suggestion is invalid and changes the flow. Also, why do we need the `filter()` with no attributes?

```suggestion
        models = self.dataset.model_set  # might need `.all()` here
        if not is_for_rdf_export:
            models = models.filter(metadata_version=self.metadata_version)
```

vitrina/datasets/views.py
Struchid: What's the point of a bare except statement here? `get_or_404` will be more than enough.

> Also you can remove later code in this view that retrieves metadata_version and use self.metadata_version

Ambiguous comment, and why does this suggestion imply that setting `self.metadata_version` inside of a views GET method a good practice? It is a request handling method, not a state container and setting the self.metadata_version and using it elsewhere expecting it to be set is bound to bear bad news in the future.

I do not understand how is this suggestion better than what's already implemented, make sure to write an explanation to your suggestions

vitrina/datasets/views.py
Struchid: If this suggestion comes from these assumptions:
- https://github.com/atviriduomenys/katalogas/pull/2233#discussion_r2697379756

Nope.

vitrina/resources/models.py
martynasracct: Makes sense. I think you can keep it in that case 🤷‍♂️

vitrina/structure/migrations/0012_populate_version_of_new_tables.py
martynasracct: All good then

vitrina/structure/migrations/0013_version_number_assigning.py
martynasracct: All good then

vitrina/resources/models.py
Gustascct: Yes there is, because an unasigned DatasetDistribution does not have Metadata, therefore no version.

vitrina/datasets/views.py
gts446: Depends wether we want to show 404 for non existing version or default to existing one. I think defaulting is better, but I guess thats opinionated.
Regarding `self.metadata_version` I see no point in additional querying of `metadata_version`, because its not going to change throughout the view. Save it to the view on first occurence if its being used later on in view, or save it to variable if its not. I don't understand why thats not a good practice?

vitrina/resources/migrations/0012_datasetdistribution_access_datasetdistribution_level_and_more.py
Gustascct: Fetched all metadata rows for datasetdistributions, created a dictionary out of them and accessed values as needed. By my understanding it should make it better

vitrina/datasets/templates/vitrina/datasets/detail.html
Gustascct: Changed

vitrina/resources/models.py
Gustascct: Added

vitrina/resources/models.py
Gustascct: Removed

vitrina/resources/models.py
Gustascct: This part was removed

vitrina/resources/models.py
Gustascct: updated

vitrina/resources/models.py
Gustascct: changed

vitrina/resources/models.py
Gustascct: renamed

vitrina/resources/models.py
Gustascct: added

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
Gustascct: yep good catch, unindented

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
Gustascct: Removed

vitrina/structure/views.py
Gustascct: changed

vitrina/resources/views.py
Gustascct: extracted

vitrina/structure/models.py
Gustascct: added

vitrina/structure/forms.py
Gustascct: Yeah not sure, but when I presented it there were no objections. Will change if people start objecting if thats okay with you

Struchid: Agreed, let's leave it as is then

vitrina/structure/models.py
Gustascct: moved

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
Gustascct: Theres nothing to suggest that it is deprecated, why do you think that this is a problem?

vitrina/datasets/views.py
Gustascct: Added the check, metadata_version is contained in a singular function and it comes from a GET request, so its safer to keep it only on the request

vitrina/datasets/views.py
Gustascct: The metadata_version comes from a GET request so it is safer to keep it as a parameter only on the same request. The get method only handles redirecting, get_context_data is the only function that uses metadata_version so there is no reason to make them depend on each other.

vitrina/datasets/services.py
Gustascct: Changed

vitrina/resources/views.py
Gustascct: changed

vitrina/resources/views.py
Gustascct: removed

vitrina/resources/models.py
Gustascct: renamed

vitrina/structure/models.py
Gustascct: This method ensures that the saving of a model and creation of needed datasetdistribution metadata is correct.

There are 3 ways to create a model, through a form, through a DSA import and one by publishing a model (it creates a duplicate). Having all of the logic in the save method saves me from having to duplicate the code in different paths and from having to customize it to each way of creation.

Also updating the models datasetdistribution should also run these checks, which do run if the code is in the save method. So in all, at least 4 distinct cases for when this code should run. 

I can explain the logic in a call.

Yes, there is a business requirement for it being created this way.

vitrina/structure/models.py
Gustascct: I agree that this is not ideal but having the connection to Version only through Metadata is also not the best idea. There should be a table between Dataset and Metadata. Wont touch now, but an improvement can be made

vitrina/resources/migrations/0012_datasetdistribution_access_datasetdistribution_level_and_more.py
Gustascct: DatasetDistribution does have this field already, but select_related is incorrect and prefetch_related works in shell but does not work in the migration for some reason. Either way made a speed improvement

vitrina/structure/migrations/0011_create_version_for_datasets_with_no_version.py
gts446: Sorry, I mixed up `Dataset.version` with `Dataset.dataset_version`

vitrina/structure/models.py
gts446: Another thing. Shouldn't there be some exception raised or something on line 328 instead of empty return? I understand if we run into such case then it means serious problems, because data is corrupt, no ?

vitrina/structure/models.py
gts446: Then maybe add extra method(s) that would implement this logic and call it explicitly where needed before `save`? For me whole logic inside `save` looks scary and confusing, and the fact that its inside `save` makes it even scarier :). Maybe a 3rd opinion would be good here. @lescijuscct ?

vitrina/structure/models.py
Gustascct: No, this happens when publishing a version. At one moment there is a DatasetDistribution which does not match the version of the model, this has to be let through here. There are validations higher up that make sure that this does not cause a problem

vitrina/structure/models.py
Gustascct: Asked for input from Gytis, Martynas and Darius, they agreed that having it in the save makes sense

vitrina/resources/views.py
Gustascct: has_permission should run first, but it is not guaranteed, so setting self.resource in it would be incorrect because if it does not run the view will fail. Solving this problem would require adding a few functions which would have to be called from all 3 functions which would defeat the purpose.

gts446: Could you give example when `has_permission` would not run first? I might be missing something.

Gustascct: I do not know but I think it is reasonable to have such a doubt. 

In addition, has_permissions is not the right place to define self variables

gts446: dispatch chain is `<..>->dispatch()->has_permission()-><..>` (Thats because this view inherits `PermisisonRequiredMixin`) if user has permission then it goes to whatever http method is and remaining logic.
If not in `has_permission` then you can declare it inside `dispatch`, or `setup` (which is run before dispatch), up to you. But either way `get_object_or_404(DatasetDistribution, id=self.kwargs["pk"])` should not be called three times in a single view.

Struchid: There is no case at the top of my head, where it would be a good idea to set instance attributes inside anything else than the init method of a class (`setup()` for `TemplateView` classes), we do not want to bring in that kind of burden in the case that something is run before the methods like your suggested `dispatch()` is, even marked on your example that somethings are run before `dispatch()`.

Anyhow, this is not a breaking change and a lot of this is built atop of old logic, so refactoring every single place while only wanting to add this kind of minimal improvement is a big burden to the developer (Gustas in this case) and the tasks would grow massively. To avoid this, we could add a ticket for improvements and let the PO decide on how much this kind of an improvement would be a priority.

@Gustascct ultimately it is for you to decide, if you want to move the resource to instance var and to the `setup()` method and improve this class or you're just building atop of this already re-use full class. If not - resolve the issue.

The last two things — regarding code reviews. Threads like these are slow and can not be found easily enough and it's better to discuss it synchronically (either by a call or a talk live). The other thing - it is not nice to block fellow developers with `Request for changes` just to bring in these kind of improvements, so let's avoid using that button in the future and use it only on changes, that would most definitely break production once deployed.

gts446: `__init__` is definitely not the right place to declare view attributes.

PO has already mentioned that there won't be separate tasks for refactoring/old code improvements, but that could/should be done together with other tasks where implementation "touches" the code that needs optimization/refactoring.

This change is 1 minute of coding, and it forms a good practice. So not sure why this discussion is where it is...

Regarding `Request for change` has already been discussed with @lescijuscct

gts446 (commented):
.

gts446 (approved):

vitrina/structure/migrations/0013_dataset_datasetdistribution_data_fix.py
Gustascct: Yes, but the logic was approved by Mantas

vitrina/resources/models.py
Gustascct: This part was taken from a previous implementation as discussed before so left it as is. Improved the naming

PR #2235, opened 2025-12-31
Ft/versionable tables tests fix

Gustascct:
DO NOT REVIEW `tests/structure/test_views.py`, the diff comes from factory changes. Most of the diff also comes from this file, so the real size of PR is way smaller

Changes: 

- Added .follow() in places where the main Dataset page is opened, because it now contains a redirect.
- DatasetFactory changed to automatically create a Metadata instance (always happens in real life)
- Conftests changed to account for the DatasetFactory change.
- Changed urls so that they are versioned.
- Changed factory usage so that models are created with versions.
- Deleted useless tests.


vitrina/datasets/factories.py
```diff
@@ -125,6 +126,16 @@ def category(self, create, extracted, **kwargs):
             for category in extracted:
                 self.category.add(category)
 
+    @factory.post_generation
+    def metadata(self, create, extracted, **kwargs):
```
Struchid: Let's keep consistent and add typehints ⭐

tests/comments/test_views.py
```diff
@@ -38,10 +38,10 @@ def test_comment_is_not_public_user_staff(app: DjangoTestApp):
     dataset = DatasetFactory()
     ct = ContentType.objects.get_for_model(dataset)
     app.set_user(user)
-    form = app.get(dataset.get_absolute_url()).forms["comment-form"]
+    form = app.get(dataset.get_absolute_url()).follow().forms["comment-form"]
     form["is_public"] = False
     form["body"] = "Test comment"
-    resp = form.submit().follow()
+    resp = form.submit().follow().follow()
```
Struchid: Let's add a comment explaining why

tests/datasets/test_views.py
```diff
@@ -4224,13 +4234,12 @@ def test_dataset_rdf_download__dataset_with_spinta_data(app: DjangoTestApp):
...
     )
-    model = ModelFactory(dataset=dataset, distribution=dist)
-    MetadataFactory(
-        content_type=ContentType.objects.get_for_model(model),
-        object_id=model.pk,
-        dataset=dataset,
-        name="test/dataset/TestModel",
-    )
+    model = ModelFactory(dataset=dataset, metadata_version=dataset.metadata.first().metadata_version)
+    MetadataFactory(dataset=dataset,
```
Struchid: Is github displaying this wrong or is there a problem with indentations here?

tests/resources/test_views.py
```diff
@@ -278,26 +349,47 @@ def test_create_resource_with_existing_download_url(app: DjangoTestApp):
...
+
     user = UserFactory()
     app.set_user(user)
-    response = app.get(reverse('resource-detail', args=[dataset.pk, resource.pk]), expect_errors=True)
+    response = app.get(resource_detail_url, expect_errors=True)
     assert response.status_code == 403
 
 
 @pytest.mark.django_db
 @pytest.mark.parametrize(
```
Struchid, line 400: You could 2x mark.parametrize here to not repeat all the patterns

tests/resources/test_views.py
```diff
@@ -307,139 +399,194 @@ def test_distribution_detail_with_non_public_dataset_with_access(app: DjangoTest
...
 
 
 @pytest.mark.django_db
 def test_distribution_detail_dynamic_resource_json(app: DjangoTestApp):
     dataset = DatasetFactory(is_public=True)
-    resource = DatasetDistributionFactory( uapi_format=True)
+    metadata_version = dataset.metadata.first().metadata_version
+    resource = DatasetDistributionFactory(dataset=dataset, uapi_format=True)
+    model = ModelFactory(dataset=dataset, metadata_version=metadata_version)
+    model_metadata = MetadataFactory(dataset=dataset,
```
Struchid: Check for formatting problems here as well

tests/uapi/dataset/test_views.py
```diff
@@ -348,7 +347,7 @@ def test_serialization_validation_error(
 
         assert response.status_code == status.HTTP_400_BAD_REQUEST
         assert Dataset.objects.filter(organization=organization).count() == 0
-        assert Metadata.objects.count() == 0
+        assert Metadata.objects.count() - 1 == 1
```
Struchid: Let's filter instead of minus one

tests/structure/test_views.py
```diff
@@ -904,35 +972,31 @@ def test_getall(app: DjangoTestApp):
...
-    MetadataFactory(
-        content_type=ContentType.objects.get_for_model(prop_2),
-        object_id=prop_2.pk,
-        dataset=dataset,
-        name='prop_2',
-        type='integer'
+        metadata_version=version,
     )
 
     with patch('vitrina.structure.services.requests.get') as mock_get:
```
Struchid, line 1003: Check if this patch does anything

tests/structure/test_views.py
```diff
@@ -1063,35 +1123,31 @@ def test_getone(app: DjangoTestApp):
...
```
Struchid, line 1154: And other places:D

tests/structure/test_views.py
Gustascct, line 1: Do not review, all of the diff comes from updating Factories with version

tests/structure/test_views.py
```diff
@@ -4533,118 +4597,57 @@ def test_draft_metadata_form_does_not_change_status_is_kept(app: DjangoTestApp):
...
-    resp_models = app.get(reverse("model-structure", args=[structure.dataset.pk, "Country"]))
-    assert list(resp_models.context["models"].values_list("metadata__status__codename", flat=True)) == ["completed"]
-
-    resp_props = app.get(reverse("property-structure", args=[structure.dataset.pk, "Country", "administration"]))
-    assert resp_props.context["prop"].metadata.get().status.codename == "completed"
-
-    resp_props = app.get(reverse("property-structure", args=[structure.dataset.pk, "Country", "administration"]))
+    resp_props = app.get(reverse("property-structure", args=[structure.dataset.pk, version.pk, "Country", "administration"]))
     prop = resp_props.context["prop"]
-    #TODO the status of enum should also be completed but because of a bug the name of the enum is changed even though nothing is submited. Change after bug fix
```
Struchid: Check if it's ok to delete this

tests/structure/test_views.py
```diff
@@ -5028,7 +5113,998 @@ def test_publish_form_shows_all_metadata_rows_denorm_props(app: DjangoTestApp):
...
+
+def test_publishing_dataset_duplicates_metadata_but_not_dataset(app: DjangoTestApp):
+    user = UserFactory(is_staff=True)
+    app.set_user(user)
+    dataset = DatasetFactory()
+    version = dataset.metadata.first().metadata_version
+    dataset_meta = dataset.metadata.first()
+
+    assert Metadata.objects.filter(dataset=dataset).count() == 1
+    assert Dataset.objects.count() - 1 == 1
```
Struchid: Filter instead of `count - 1`

tests/structure/test_views.py
```diff
@@ -5028,7 +5113,998 @@ def test_publish_form_shows_all_metadata_rows_denorm_props(app: DjangoTestApp):
...
+    prop_meta = MetadataFactory(
+        content_type=ContentType.objects.get_for_model(prop),
+        object_id=prop.pk,
+        dataset=dataset,
+        name='prop',
+        type='string',
+        metadata_version=version
+    )
+
+    assert Metadata.objects.filter(dataset=dataset).count() == 3
```
Struchid: Consider moving the initial count to a variable (if later some more are added, you would just have to change a single place instead of multiple places

Struchid (approved):
Less context in the comments due to the review done side-by-side live, other than that looks good.

vitrina/structure/views.py
Gustascct, line 1: Never used so removed

vitrina/structure/services.py
Gustascct, line 1: Removed the default creation of Saugykla API since it is incorrect

vitrina/datasets/factories.py
Gustascct: Added

tests/comments/test_views.py
Gustascct: Added

tests/datasets/test_views.py
Gustascct: Problem with indentation, fixed

tests/resources/test_views.py
Gustascct: Implemented

tests/resources/test_views.py
Gustascct: Checked, corrected

tests/uapi/dataset/test_views.py
Gustascct: Added

tests/structure/test_views.py
Gustascct: Readded

tests/structure/test_views.py
Gustascct: Nothing to filter by, one Dataset created by a migration, changed assert to == 2 and wrote a comment explaining

tests/structure/test_views.py
Gustascct: Implemented where the counts differ by * 2

tests/structure/test_views.py
Gustascct: Does not seem to be the same issue we discussed about another patch, so left it, it makes some sense being here

tests/structure/test_views.py
Gustascct: Does not seem to be the same issue we discussed about another patch, so left it, it makes some sense being here

MindaugasStanionis (approved):

wisecrow (approved):

martynasracct (approved):

PR #2253, opened 2026-01-08
Make `LOGGING` configurable

kastger:
If we had this before the issue with server space filling up from logs would've been a 1 minute fix:

```
DJANGO_LOGGING='{"version": 1, "disable_existing_loggers": false, "handlers": {"console": {"level": "INFO", "class": "logging.StreamHandler"}}, "loggers": {"elasticsearch": {"level": "WARNING", "handlers": ["console"], "propagate": false}, "spinta": {"level": "WARNING", "handlers": ["console"], "propagate": false}}, "root": {"handlers": ["console"], "level": "WARNING"}}'
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,6 +11,7 @@ SPINTA_SERVER_URL=https://127.0.0.1:8000
 SPINTA_SERVER_NAME=127.0.0.1:8000
 SPINTA_PATH=spinta-env/data
 USE_OTP_VALIDATION=True
+DJANGO_LOGGING={"version": 1, "disable_existing_loggers": false, "handlers": {"console": {"level": "DEBUG", "class": "logging.StreamHandler"}}, "root": {"handlers": ["console"], "level": "WARNING"}}
 
 RECAPTCHA_SILENCE_KEY_ERROR=True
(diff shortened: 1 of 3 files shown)


martynasracct (approved):
Honestly, I would just update `LOGGING` configuration to the one in pull request description without introducing `env`.

This issue should also be fixed from Spinta side by https://github.com/atviriduomenys/spinta/pull/1663. By only enabling logging when using Spinta as a service or Spinta as a cli tool.

karina-klinkeviciute (commented):
I would suggest to only have debug levels in the environment variables:

- Logging is code, it has a certain syntax, and we have tools in place to check it in the code. I mean, there is a python dict, which is checked for syntax, and which is not checked in the .env file. 

- developers should have control over it.

- Because it's code, it should go through all the code cycles - review, test, etc. 

- It's more transparent this way, and even if developers don't know what levels are used, they see the other configuration better, and can have better ideas on what loggers to use if there are more than 1 for example. 

Also, if we change something, for example, add a new logger, like sentry, graylog, kibana, etc, everyone can see it in the settings file and can fix the problems or know where the logs go.

PR #2290, opened 2026-01-19
1656 Make dataset.uuid required with default. Change it to UUIDField

martynasracct:
**Summary:**
Currently not all datasets have uuid set. We need `Dataset.uuid` for UAPI endpoints. This PR:
- Backfills all `Dataset.uuid` with unique UUIDs;
- Makes `Dataset.uuid` required and adds default;
- Changes `Dataset.uuid` to `UUIDField` for better UUID validation;
- Adds read-only `Dataset.uuid` to admin.

Here is the `sqlmigrate`. Looks like changing `CharField` into `UUIDField` drops one index and changes column type:
```sql
BEGIN;
--
-- Raw Python operation
--
-- THIS OPERATION CANNOT BE WRITTEN AS SQL
--
-- Alter field uuid on dataset
--
DROP INDEX IF EXISTS "dataset_uuid_356083f1_like";
ALTER TABLE "dataset" ALTER COLUMN "uuid" TYPE uuid USING "uuid"::uuid, ALTER COLUMN "uuid" SET DEFAULT '5bfe7909-ffa7-4323-acb6-091ae71624cc'::uuid;
UPDATE "dataset" SET "uuid" = '0293c91d-6cb5-47bd-b7ef-54c40af3df78'::uuid WHERE "uuid" IS NULL; SET CONSTRAINTS ALL IMMEDIATE;
ALTER TABLE "dataset" ALTER COLUMN "uuid" SET NOT NULL;
ALTER TABLE "dataset" ALTER COLUMN "uuid" DROP DEFAULT;
COMMIT;
```

**Ticket:**
https://github.com/atviriduomenys/spinta/issues/1656


wisecrow (approved):
LGTM. It would be good to have at least one test confirming uuid is generated when Dataset is created.

vitrina/datasets/migrations/0037_alter_dataset_uuid.py
```diff
@@ -0,0 +1,30 @@
...
+from django.db import migrations, models
+import uuid
+
+
+def fill_dataset_uuids(apps, schema_editor):
+    Dataset = apps.get_model("vitrina_datasets", "Dataset")
+
+    datasets_without_uuid = Dataset.objects.filter(uuid__isnull=True)
+    for dataset in datasets_without_uuid:
+        dataset.uuid = str(uuid.uuid4())
```
gts446, line 12: I don't think you need to wrap it with `str` since field itself is `UUIDField`

gts446 (approved):

vitrina/datasets/migrations/0037_alter_dataset_uuid.py
gts446: okay, I guess its because you first apply data migration for `Charfield` and only then change field type. Can't it be done in reverse? First change field type then fill data?

vitrina/datasets/models.py
```diff
@@ -247,7 +248,7 @@ class Dataset(Resource):
         blank=False,
         null=True,
     )  # TODO: Deprecated, slugs are formed from id's
-    uuid = models.CharField(unique=True, max_length=36, blank=True, null=True)  # TODO: Remove blank and null
+    uuid = models.UUIDField(unique=True, default=uuid4, editable=False)
```
Struchid, line 251: Should we define this field as an index? Won't it slow down the app if we don't?

Struchid (approved):

vitrina/datasets/migrations/0037_alter_dataset_uuid.py
martynasracct: That wouldn't change anything. What problem are you trying to solve?

vitrina/datasets/models.py
martynasracct: `unique=True` automatically adds index: https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.unique

PR #2295, opened 2026-01-20
705 delete dsa button

### LINKED ISSUE #705: Nėra galimybės ištrinti modelio/ ADSA
Per klaidą įkėlus ne tą ADSA/ modelį ar norint iš esmės jį tikslinti, nėra galimybės pašalinti pirminio įkelto ADSA/ modelio,
Siūlymas: sukurti tokį funkcionalumą, pvz. mygtukas "Pašalinti".
https://data.gov.lt/datasets/673/models/Moketojas/ 

image

VDA netyčia buvo pridėtas modelis [Humanitarinepagalba] ir jo nėra galimybės pašalinti (rodo tik kai prisijungi prie ADP superuserio ir šios įstaigos koordinatoriaus/ tvarkytojo teisėmis)
https://data.gov.lt/datasets/2372/models/

kastger:
image
image

LOOKS SIMILAR TO COMMENT DELETE MODAL (consistent desgin, not sure if there are other)
image


tests/structure/test_views.py
```diff
@@ -5032,3 +5032,94 @@ def test_publish_form_shows_all_metadata_rows_denorm_props(app: DjangoTestApp):
...
+        model = ModelFactory(dataset=dataset)
+        MetadataFactory(
+            dataset=dataset,
+            content_type=ContentType.objects.get_for_model(Model),
+            object_id=model.pk,
+            name=f"{dataset}/{model.pk}/TestModel",
+        )
+
+        resp = app.get(reverse('model-delete', args=[dataset.pk, 'TestModel']), expect_errors=True)
+        assert resp.status_code == 405
```
Struchid: Missing empty line at the eof

vitrina/structure/templates/vitrina/structure/dataset_structure.html
Struchid, line 1: Not sure how it is in other places but does the button (delete - trash bin) and the modal conform to the current style in Katalogas? If not, maybe it would be a good idea to stay consistent and keep it the way it is?

vitrina/structure/views.py
```diff
@@ -2011,6 +2011,34 @@ def get_form_kwargs(self):
         return kwargs
 
 
+class ModelDeleteView(LoginRequiredMixin, View):
+    http_method_names = ["post"]
+
+    def post(self, request, pk, model):
+        dataset = get_object_or_404(Dataset, pk=pk)
+        if not has_perm(request.user, Action.STRUCTURE, Dataset, dataset):
+            return JsonResponse({"error": "Permission denied"}, status=403)
```
Struchid, line 2020: Wouldn't Django handle this differently and raise `PermissionDenied` exception? (`from django.core.exceptions import PermissionDenied`)

vitrina/structure/views.py
```diff
@@ -2011,6 +2011,34 @@ def get_form_kwargs(self):
...
+                model_name=Func(
+                    F("metadata__name"),
+                    Value("/"),
+                    Value(-1),
+                    function="split_part",
+                    output_field=TextField(),
+                )
+            )
+            .filter(model_name=model, dataset=dataset)
+            .first()
```
Struchid: I would do `get_object_or_404` instead of `first()` due to the possible duplicates here

Struchid (approved):

vitrina/structure/templates/vitrina/structure/dataset_structure.html
kastger: Other place I found was comments deletion modal (added screenshot to this PR description). Made this modal similar to comment modal. Since we don't have design doc (or basically any docs for devs) its a wild west. Good point tho.

vitrina/structure/views.py
kastger: You are right. I did reuse this from the `ModelUpdateView`, but your suggestion is better.

vitrina/structure/views.py
kastger: PermissionDenied would return HTML instead of JSON, Since I am using AJAX here handling HTML would add extra complexity for same functionality. I like to KISS.

Also, I think this follows the same pattern as comments.

gts446:
I have only seen `confirm_delete.html` used as template for delete confirmation pages. Is this JS modal replacement or legacy?

wisecrow (approved):

kastger:
> I have only seen `confirm_delete.html` used as template for delete confirmation pages. Is this JS modal replacement or legacy?

@gts446 I think these are for different purposes - `confirm_delete.html` is used to delete single item such as `Dataset`, and here in this PR we have multiple `Model` within one `Dataset` and if we need to delete multiple `Models` would lead to clunky UX because of redirecting back and forth for the user.

martynasracct (approved):

PR #2300, opened 2026-01-21
[1647] Add API conforming to UAPI for Agent

Struchid:
Agent API endpoint for retrieving data of a specific agent, needed to retrieve information about agent from Spinta during synchronization process.

Only looks big due to a single renamed file, I marked which files to ignore and why to ignore them with my own review comments.

**UAPI Reference:**
- https://ivpk.github.io/uapi/

### DIFF
--- a/tests/uapi/agent/test_template_views.py
+++ b/tests/uapi/agent/test_template_views.py
@@ -1,0 +1,293 @@
+import secrets
+from http import HTTPStatus
+from unittest.mock import patch
+
+import pytest
+from django.urls import reverse
+from django_webtest import DjangoTestApp
+from django.contrib.contenttypes.models import ContentType
+from vitrina.api.models import ApiKey
+from vitrina.datasets.factories import DatasetFactory, AgentFactory
+from vitrina.datasets.models import Dataset, DCATResourceSubclass
+from vitrina.structure.factories import MetadataFactory
+from vitrina.uapi import AgentType, Environment
+from vitrina.orgs.factories import OrganizationFactory
+from vitrina.orgs.models import Organization
+from vitrina.uapi.models import Agent, RequestHistory
+from vitrina.orgs.services import hash_api_key
+from vitrina.users.models import User
+
+
+pytestmark = pytest.mark.django_db
+
+
+class TestAgentList:
+    def test_success(self, app: DjangoTestApp, representative_user: User, organization: Organization):
+        app.set_user(representative_user)
+
+        another_organization = OrganizationFactory()
+        dataset = DatasetFactory(service=True, organization=organization)
+        agent_1 = Agent.objects.create(title="Agent 1", organization=organization, service=dataset)
+        agent_2 = Agent.objects.create(title="Agent 2", organization=organization, service=dataset)
+        archived_agent = Agent.objects.create(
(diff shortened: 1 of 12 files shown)


tests/uapi/agent/test_template_views.py
Struchid, line 1: Ignore, just renamed the file, Github seems to think that I've added all of this instead.

tests/uapi/agent/test_views.py
Struchid, line 1: This also is a problem here, due to the renaming, it thinks that this logic is changed and not added anew. Can ignore, nothing special, just simple test cases for the added endpoint.

MindaugasStanionis (approved):
LGTM

vitrina/uapi/views/views.py
```diff
@@ -373,3 +375,25 @@ def list(self, request: Request, *args: Any, **kwargs: Any) -> Response:
             _type=extract_type_from_url(self.request.build_absolute_uri()),
         )
         return Response(serializer.data, status=status.HTTP_200_OK)
+
+
+class AgentViewSet(UAPIExceptionHandlerMixin, AgentAuthViewSetMixin, viewsets.ModelViewSet):
+    required_scopes = {"list": ["uapi:/datasets/gov/vssa/dcat/Agent/:getall"]}
```
Gustascct: Why not import the ModelViewSet so that viewsets. is not needed?

Gustascct (approved):

vitrina/uapi/views/views.py
Struchid: No reason, does not matter much IMO in this case, but here you go:D [32c26e7](https://github.com/atviriduomenys/katalogas/pull/2300/commits/32c26e7856f18e6eea398dae8566eb8f051fd42f)

vitrina/uapi/serializers/uapi_serializers.py
```diff
@@ -41,6 +41,13 @@ def to_representation(self, instance: Model) -> dict:
     def get__type(self, obj: Model) -> str:
         return (self.context.get("_type", "") or "").removeprefix(TYPE_PREFIX_TO_REMOVE)
 
+    def get__id(self, obj: Model) -> str:
+        if hasattr(obj, "id"):
+            return str(obj.id)
+        elif hasattr(obj, "uuid"):
+            return str(obj.uuid)
+        return ""
```
martynasracct: In my branch I chose to implement `BaseUUIDObjectMixin` instead of doing it this way, because if model have both `id` and `uuid` fields - `id` will be used by default.

On the other hand, maybe it's better to have this just as a backup, and only use `BaseUUIDObjectMixin` if you specifically need `uuid` and not `id`, when both available

martynasracct (approved):

vitrina/uapi/serializers/uapi_serializers.py
Struchid: I agree with your first suggestion, I'll leave it consistent with your changes and we can bring in that serializer method field if needed in the future. Added in [46d9e63](https://github.com/atviriduomenys/katalogas/pull/2300/commits/46d9e6342d5c81197c8c11cd3ed2e6ff3029ce23)

PR #2313, opened 2026-01-22
Added the possibility to disable publish button

Gustascct:
Introduced django-flags which lets us create a flag in settings and control it through admin. The default value is set as False

**Migrations are out of scope, just forgot to run them before**


vitrina/settings.py
```diff
@@ -556,3 +557,5 @@
 #   "vitrina.users.models.*"          → excludes all models in that module
 #   "vitrina.comments.models.Comment" → excludes a specific model
 NOT_VERSIONED_MODELS = {}
+
+IS_PUBLISH_BUTTON_ACTIVE = False
```
Struchid: This should not be a setting at settings.py but rather something we could enable/disable in the Django Admin

vitrina/resources/migrations/0012_datasetdistribution_access_datasetdistribution_level_and_more.py
```diff
@@ -40,7 +40,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='datasetdistribution',
             name='access',
-            field=models.IntegerField(blank=True, choices=[(0, 'nepasirinkta'), (1, 'private'), (2, 'protected'), (3, 'public'), (4, 'open')], null=True, verbose_name='Prieiga'),
+            field=models.IntegerField(blank=True,choices=[(0, 'private'), (1, 'protected'), (2, 'public'), (3, 'open')], null=True,verbose_name='Prieiga'),
```
MindaugasStanionis, line 43: Just checking - are the migration changes part of this ticket’s scope? I understood the goal was to disable a button.

MindaugasStanionis (approved):

vitrina/resources/migrations/0012_datasetdistribution_access_datasetdistribution_level_and_more.py
Gustascct: No, I forgot to run migrations and two appeared so just merged them into old ones.

vitrina/structure/migrations/0013_dataset_datasetdistribution_data_fix.py
Struchid, line 1: Make sure that these migrations are not already deployed to other environments, if they are, create new ones instead of adjusting old ones

vitrina/settings.py
```diff
@@ -556,3 +557,8 @@
 #   "vitrina.users.models.*"          → excludes all models in that module
 #   "vitrina.comments.models.Comment" → excludes a specific model
 NOT_VERSIONED_MODELS = {}
+FLAGS = {
+    "publish_button": [
+        {"condition": "boolean", "value": True},
```
Struchid: If we are introducing this kind of flag I'd assume that we should set the initial value to `False`, otherwise I don't really get why we need a flag at all - to disable it if it's not working correctly?

Struchid (approved):

vitrina/structure/migrations/0013_dataset_datasetdistribution_data_fix.py
Gustascct: Yeah I created them in other unmerged PR's

vitrina/settings.py
Gustascct: The tests pass with True, the value will have to be changed to False in the environment

vitrina/structure/views.py
```diff
@@ -3493,6 +3495,8 @@ class PublishVersionView(PermissionRequiredMixin, CreateView):
     def dispatch(self, request, *args, **kwargs):
         self.dataset = get_object_or_404(Dataset, pk=kwargs.get("pk"))
         self.metadata_version = get_object_or_404(_Version, pk=kwargs.get("version_id"))
+        if not flag_enabled("publish_button", request=request):
```
wisecrow: Let's provide a meaningful message to the user why user was redirected here.

wisecrow (approved):
I suggest to check feature is enabled when url for publishing is defined in urls.py.
In other case, url is defined but does not work.

vitrina/settings.py
Gustascct: Changed it to False and added a conftest that autoruns for each test and sets this value to True. Sorry for the confusion

vitrina/structure/views.py
Gustascct: Talked in person and decided to remove the url altogether. This is done by setting a decorator on PublishVersionView

PR #2327, opened 2026-02-03
2326 Encode TAB space in RDF export

### LINKED ISSUE #2326: TAB space užkodavimas
Gautas pastebėjimas iš Europos duomenų portalo dėl bandymo harvestinti restricted duomenis iš data.gov.lt.

Our technical team has been trying to launch the harvesting process. Unfortunately, they inform us that the file has an issue with a literal on line 1363, which contains TAB characters, which is not allowed in XML. 
If you need to keep the TABs, they must be encoded as follows:

Problema yra paliktos tuščios paliktos eilutės tekste. Pvz rinkinys https://data.gov.lt/datasets/4034/resource/17996

Image

Reikia padaryti kodavimą kaip siūloma Europos duomenų portalo pasiūlyme.

kastger:
Laiškas iš EDP:

Our technical team has been trying to launch the harvesting process. Unfortunately, they inform us that the file has an issue with a literal on line 1363, which contains TAB characters, which is not allowed in XML.
If you need to keep the TABs, they must be encoded as follows:
```
&#x9;
```

If questions, please let us know. We can also set up a brief call to solve the issue asap. If so, please provide us a couple of slots and we will confirm availability.

### DIFF
--- a/tests/api/test_views.py
+++ b/tests/api/test_views.py
@@ -33,6 +33,8 @@ from vitrina.classifiers.factories import FrequencyFactory
 from vitrina.resources.factories import FileFormat
 from vitrina.utils import RevisionComment, RevisionSource
 
+from vitrina.api.helpers import _encode_xml_control_chars
+
 
 @pytest.mark.django_db
 def test_retrieve_catalog_list_without_api_key(app: DjangoTestApp):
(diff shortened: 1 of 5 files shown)


kastger:
PR is ready, I have generated new `.rdf` file and asked Aurimas to send it to EDP for inspection. Will wait for their confirmation before moving forward with this PR since they might find other issues.

tests/api/test_views.py
```diff
@@ -2217,3 +2219,64 @@ def test_edp_dcat_ap_rdf_hvd_dataset(app: DjangoTestApp):
     </dcat:Dataset>
 </rdf:RDF>"""
     )
+
+
+class EncodeXmlControlCharsTests(TestCase):
```
martynasracct: We should not use Django `TestCase`, since we are using `pytest` for tests.

tests/api/test_views.py
```diff
@@ -2217,3 +2219,64 @@ def test_edp_dcat_ap_rdf_hvd_dataset(app: DjangoTestApp):
...
+    def test_encode_xml_control_chars_no_tabs(self):
+        result = _encode_xml_control_chars("No tabs here")
+        self.assertEqual(result, "No tabs here")
+
+    def test_encode_xml_control_chars_empty_string(self):
+        result = _encode_xml_control_chars("")
+        self.assertEqual(result, "")
+
+
+class EdpDcatApRdfTabEncodingTests(TestCase):
```
martynasracct: Same for this class

martynasracct (approved):

tests/api/test_views.py
kastger: Refactored to not use classes at all here and instead parametrize cases to make the tests shorter

Struchid (approved):

PR #2333, opened 2026-02-05
2332 fix model data table

### LINKED ISSUE #2332: Model data table not displaying data
Aurimas informed that despite correct `saugykla` url the table is not being displayed:
https://data.gov.lt/datasets/2088/versions/1605/data/JuridinisAsmuo/
Image

The problem is most likely related to the recent changes done with versioning #2233 :
Image

The request for the table returns 404 status code, because the url was changed in django:
```
Before:
datasets/<int:pk>/data/<str:model>/table/

After:
datasets/<int:pk>/versions/<int:version_id>/data/<str:model>/table/
```

But it was not changed in the AJAX requests in `vitrina/structure/templates/vitrina/structure/model_data.html`:
Image

**Goal:** Use correct url for datatable retrieval. Bonus: Add integration test that would cover such case.

kastger:
Details are in the #2332 task description
image

### DIFF
--- a/tests/structure/test_views.py
+++ b/tests/structure/test_views.py
@@ -319,6 +319,54 @@ def test_model_data_with_string_operators(app: DjangoTestApp, operator: str):
     assert resp.context["selected_cols"] == ["prop_1"]
 
 
+def test_model_data_page_contains_correct_table_url(client):
+    user = UserFactory(is_staff=True)
+    client.force_login(user)
+
+    version = VersionFactory()
+    model = ModelFactory(dataset=version.dataset, metadata_version=version)
+    dataset = version.dataset
+
+    MetadataFactory(
+        content_type=ContentType.objects.get_for_model(model),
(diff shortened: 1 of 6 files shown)


martynasracct (approved):

Struchid (approved):

tests/structure/test_views.py
```diff
@@ -319,6 +319,54 @@ def test_model_data_with_string_operators(app: DjangoTestApp, operator: str):
...
+    dataset = version.dataset
+
+    MetadataFactory(
+        content_type=ContentType.objects.get_for_model(model),
+        object_id=model.pk,
+        dataset=dataset,
+        name="test/dataset/TestModel",
+        metadata_version=version,
+    )
+    MetadataFactory(
```
wisecrow: You can use `DatasetFactory` and Metadata would be created automatically

wisecrow (approved):

PR #2408, opened 2026-02-20
refactor OrganizationBaseViewMixin

gts446:
No logic has changed.
Only refactored `OrganizationViewMixin` and fixed tabs and breadcrumbs not displaying properly in some places.
Added `parent_links` to `OrganizationViewMixin` to avoid repeating it inside every view. 
Also removed `parent_links` from templates, because it is already added to `base.html`

This change will be needed later on because I will reuse `OrganizationViewMixin` for `Agent` related views.

### DIFF
--- a/vitrina/orgs/templates/vitrina/orgs/tabs.html
+++ b/vitrina/orgs/templates/vitrina/orgs/tabs.html
@@ -30,7 +30,7 @@
                 {% translate "Panaudojimo atvejai" %}
             </a>
         </li>
-        <li class="{%if request.resolver_match.url_name in 'organization-agreement-list,organization-agreement-detail' %} is-active {% endif %}">
+        <li class="{% if 'organization-agreement' in request.resolver_match.url_name %} is-active {% endif %}">
             <a href="{% url 'organization-agreement-list' organization.id %}">
                 {% translate "Sutartys" %}
             </a>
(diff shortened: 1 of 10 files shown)


vitrina/orgs/views.py
```diff
@@ -122,11 +122,12 @@
 logger = logging.getLogger()
 
 
-class OrganizationBaseViewMixin:
+class OrganizationBaseViewMixin(PlanMixin):
```
Struchid: I would advise against creating mixin inheritance chains. It will only take a single change on PlanMixin to have to refactor all of this back to what it was again

vitrina/orgs/templates/vitrina/orgs/tabs.html
```diff
@@ -43,14 +43,14 @@
             </li>
         {% endif %}
         {% if can_view_agents %}
-            <li class="{% if request.resolver_match.url_name == 'agent-list' or request.resolver_match.url_name == 'agent-create' or request.resolver_match.url_name == 'agent-update' %} is-active {% endif %}">
+            <li class="{% if 'agent' in request.resolver_match.url_name %} is-active {% endif %}">
```
Struchid: This is too implicit, what if there is another object with `agent` in url name, that is not related to your use case? Let's not do that and explicitly define the url's we expect to be marked is-active with or else this is just a bug waiting in a queue for production

vitrina/orgs/views.py
Struchid: Why introduce even tighter coupling here? Let's not inherit mixins from other mixins. One change to `PlanMixin` and this has to be refactored back to what it was. It brings more problems than solutions

vitrina/plans/views.py
```diff
@@ -116,7 +116,7 @@ def get_context_data(self, **kwargs):
         return context
 
 
-class PlanHistoryView(PlanMixin, OrganizationBaseViewMixin, HistoryView):
+class PlanHistoryView(OrganizationBaseViewMixin, HistoryView):
```
Struchid, line 119: Let's return to two different mixins instead of one inheriting the other

vitrina/orgs/templates/vitrina/orgs/tabs.html
martynasracct: Agree with Struchid. Something like this might be more implicit

```suggestion
            <li class="{% if request.resolver_match.url_name in 'agent-list,agent-create,agent-update' %} is-active {% endif %}">
```

martynasracct (approved):

Struchid (approved):

kastger:
Also, for the next time it would be good to add an entry to `CHANGES.rst`, even if its a refactor. This codebase is held together by duct tape and prayers, so having a paper trail helps  when debugging weird issues later.

PR #2444, opened 2026-03-02
Bg/2434 missing version status

### LINKED ISSUE #2434: Kuriant duomenų išteklių per partnerių API, nesusikuria juodraščio versija
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/2401
#

Sukūrus duomenų išteklių per partnerių API, nesusikuria juodraščio versija.
Image
Taip pat, metama klaida jei norime eksportuoti neimportuotą struktūrą.

MindaugasStanionis:
Summary:

Added metadata field for dataset when creating through post request

Ticket:
https://github.com/atviriduomenys/katalogas/issues/2434

### DIFF
--- a/tests/api/test_views.py
+++ b/tests/api/test_views.py
@@ -601,17 +601,18 @@ def test_create_dataset(app: DjangoTestApp):
 
     assert dataset_objects.count() == 1
     dataset = dataset_objects.first()
+    dataset_metadata = dataset.metadata.first()
     assert dataset.language == "en lt"
     assert list(dataset.tags.all()) == ["tag1", "tag2"]
     assert dataset.frequency == frequency
     assert list(dataset.category.all()) == [category]
     assert dataset.organization == organization
     assert Version.objects.get_for_object(dataset).count() == 1
-    assert dataset.metadata.first().metadata_version.status == VersionStatus.DRAFT
-    assert dataset.metadata.first().title == "Test dataset"
-    assert dataset.metadata.first().description == "Test dataset"
-    assert dataset.metadata.first().version == 1
-    assert dataset.metadata.first().object_id == dataset.pk
+    assert dataset_metadata.metadata_version.status == VersionStatus.DRAFT
+    assert dataset_metadata.title == "Test dataset"
+    assert dataset_metadata.description == "Test dataset"
+    assert dataset_metadata.version == 1
+    assert dataset_metadata.object_id == dataset.pk
     version = Version.objects.get_for_object(dataset).select_related("revision").first()
     assert version.revision.comment == revision_comment.to_json()
     assert version.revision.user == representative.user
--- a/tests/api/test_views.py
+++ b/tests/api/test_views.py
@@ -10,6 +10,7 @@ from django.utils import timezone
 from django_webtest import DjangoTestApp
 from reversion.models import Version
 
+from vitrina.structure import VersionStatus
 from vitrina.testing.templates import strip_empty_lines
 from vitrina.api.exceptions import DuplicateAPIKeyException
 from vitrina.api.factories import APIKeyFactory
@@ -606,6 +607,7 @@ def test_create_dataset(app: DjangoTestApp):
     assert list(dataset.category.all()) == [category]
     assert dataset.organization == organization
     assert Version.objects.get_for_object(dataset).count() == 1
+    assert dataset.metadata.first().metadata_version.status == VersionStatus.DRAFT
     version = Version.objects.get_for_object(dataset).select_related("revision").first()
     assert version.revision.comment == revision_comment.to_json()
     assert version.revision.user == representative.user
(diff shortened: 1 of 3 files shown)


vitrina/api/serializers.py
```diff
@@ -242,6 +249,26 @@ def create(self, validated_data):
                 instance.category.add(category)
         for tag in keywords:
             instance.tags.add(tag)
+
+        draft_metadata_version = _Version.objects.create(
+            dataset=instance,
+            version=1,
```
Gustascct, line 257: If I understand correctly, if Version or Metadata creation fails, the Dataset instance will still live in the database. Maybe wrapping everything in transaction.atomic would be correct?

Gustascct (approved):

tests/api/test_views.py
```diff
@@ -466,9 +467,15 @@ def test_create_dataset(app: DjangoTestApp):
     assert list(dataset.category.all()) == [category]
     assert dataset.organization == organization
     assert Version.objects.get_for_object(dataset).count() == 1
+    assert dataset.metadata.first().metadata_version.status == VersionStatus.DRAFT
+    assert dataset.metadata.first().title == "Test dataset"
+    assert dataset.metadata.first().description == "Test dataset"
+    assert dataset.metadata.first().version == 1
+    assert dataset.metadata.first().object_id == dataset.pk
```
Struchid: save the dataset.metadata.first() to a variable and re-use, o use in retrieving the object more than once

vitrina/api/serializers.py
```diff
@@ -230,19 +238,42 @@ def create(self, validated_data):
...
-            instance.language = " ".join(languages)
-        if periodicity and Frequency.objects.filter(title=periodicity["title"]).exists():
-            instance.frequency = Frequency.objects.filter(title=periodicity["title"]).first()
-        if theme and Category.objects.filter(title__in=theme).exists():
-            for category in Category.objects.filter(title__in=theme):
-                instance.category.add(category)
-        for tag in keywords:
-            instance.tags.add(tag)
-        instance.save()
+        with transaction.atomic():
```
Struchid: Instead of indenting most of the logic, I'd just use the decorator on the method

vitrina/api/serializers.py
```diff
@@ -230,19 +238,42 @@ def create(self, validated_data):
...
+            instance = super().create(validated_data)
+            instance.origin = Dataset.API_ORIGIN
+            instance.organization = self.context.get("organization")
+            if languages:
+                instance.language = " ".join(languages)
```
Struchid: Not sure what you need the instance.language to be here if languages is None, but you can avoid the check (at least I think you can)

```suggestion
        instance.language = " ".join(languages) or [] #  (or None?)
```

vitrina/api/serializers.py
```diff
@@ -230,19 +238,42 @@ def create(self, validated_data):
...
-        if periodicity and Frequency.objects.filter(title=periodicity["title"]).exists():
-            instance.frequency = Frequency.objects.filter(title=periodicity["title"]).first()
-        if theme and Category.objects.filter(title__in=theme).exists():
-            for category in Category.objects.filter(title__in=theme):
-                instance.category.add(category)
-        for tag in keywords:
-            instance.tags.add(tag)
-        instance.save()
+        with transaction.atomic():
+            instance = super().create(validated_data)
```
Struchid: This line seems that it could be used at the end of this transaction and you could save values to the `validated_data` instead of instance for the correct structure to be created.

The current flow will create -> update (second query) with the data.

Might at least do this after origin, organization, languages and periodicity is set.

vitrina/api/serializers.py
```diff
@@ -230,19 +238,42 @@ def create(self, validated_data):
...
+            instance.organization = self.context.get("organization")
+            if languages:
+                instance.language = " ".join(languages)
+            if periodicity:
+                frequency = Frequency.objects.filter(title=periodicity["title"]).first()
+                if frequency:
+                    instance.frequency = frequency
+            if theme:
+                for category in Category.objects.filter(title__in=theme):
+                    instance.category.add(category)
```
Struchid: Use bulk add:


```suggestion
            categories = list(Category.objects.filter(title__in=theme))
            instance.category.add(*categories)
```

vitrina/api/serializers.py
```diff
@@ -230,19 +238,42 @@ def create(self, validated_data):
...
+            for tag in keywords:
+                instance.tags.add(tag)
```
Struchid: Same here

Struchid (approved):

vitrina/api/serializers.py
MindaugasStanionis: Fixed in 7c8b6ed7bc1bbf47b7a34b847c66b0670b346f24

vitrina/api/serializers.py
MindaugasStanionis: Added decorator in 7c8b6ed7bc1bbf47b7a34b847c66b0670b346f24

vitrina/api/serializers.py
MindaugasStanionis: Fixed in 7c8b6ed7bc1bbf47b7a34b847c66b0670b346f24

tests/api/test_views.py
MindaugasStanionis: Fixed in 7c8b6ed7bc1bbf47b7a34b847c66b0670b346f24

PR #2445, opened 2026-03-02
[2349] Base reference issues when more than one file is used

### LINKED ISSUE #2349: DSA export base model issues
**Specification:**
- https://github.com/atviriduomenys/katalogas/issues/2402

---

### Case 1:

The import for this should remain as is (base is linked correctly due to `metadata_version` being the same for model and base):

<details>
<summary>Model & Base in one file</summary>

```csv
id,dataset,resource,base,model,property,type,ref,source,prepare,level,status,visibility,access,uri,eli,title,description
,example,,,,,,,,,,,,,,,,
,,,,Animal,,,,,,0,completed,public,,,,,
,,,,,id,string,,source_animal_id,,4,completed,package,protected,,,,
,,,Animal,,,,,,,1,completed,public,,,,,
,,,,Dog,,,,,,0,completed,public,,,,,
,,,,,action,string,,source_dog_action,,4,completed,package,protected,,,,
,,,/,,,,,,,,,,,,,,
```

</details>

---

A notable issue arose while dealing with Base rows while exporting an uploaded manifest to/from Catalog.

#### Input:

<details>
<summary>File 1 (1st Dataset)</summary>

```csv
id,dataset,resource,base,model,property,type,ref,source,prepare,level,status,visibility,access,uri,eli,title,description
,example,,,,,,,,,,,,,,,,
,,,,Animal,,,,,,0,completed,public,,,,,
,,,,,id,string,,source_animal_id,,4,completed,package,protected,,,,
,,,,,,,,,,,,,,,,,
```

</details>

<details>
<summary>File 2 (2nd Dataset)</summary>

```csv
id,dataset,resource,base,model,property,type,ref,source,prepare,level,status,visibility,access,uri,eli,title,description
,example2,,,,,,,,,,,,,,,,
,,,example/Animal,,,,,,,1,completed,public,,,,,
,,,,Dog,,,,,,0,completed,public,,,,,
,,,,,action,string,,source_dog_action,,4,completed,package,protected,,,,
,,,/,,,,,,,,,,,,,,
```

</details>

#### Output:

<details>
<summary>File</summary>

```csv
id,dataset,resource,base,model,property,type,ref,source,source.type,prepare,origin,count,level,status,visibility,access,uri,eli,title,description
b788c221-47bf-4d84-8b81-fd0ac570eb1e,example2,,,,,,,,,,,,,,,,,,Pumpureliai,abc
10ed1071-2cb3-42b6-874c-be04687ba57c,,,,Dog,,,,,,,,,0,completed,public,,,,,
6740119b-bb2a-4c85-8dfe-aeb56716f7d2,,,,,action,string,,source_dog_action,,,,,4,completed,package,protected,,,,
,,,,,,,,,,,,,,,,,,,,
```

</details>

---

The problem with the output is the lost `Base` row. It is lost due to the reason of how it is linked during the import process.

During import, a globally unique metadata version is assigned to all rows imported from a single manifest. It is assumed, that the base model will be defined in the same manifest that the Base row is present. But there might be cases, where we want to import other manifests, where the Base model is re-used from an already imported Manifest.

#### Solution: 

- https://github.com/atviriduomenys/katalogas/issues/2349#issuecomment-3964485900

It should not be assumed that the base will be imported with its model in one file, but rather:

a) If in one file - correctly link with the usage of `metadata_version`.
b) Find the latest `STABLE` version of the model that the Base is referencing and use that.

Struchid:
It was assumed that the base and its model (the referenced one with `Base`) will be imported via a single file. The assumption is wrong since there might be a need to reference an old imported model (with Base).

#### Solution:

- If both the model is imported and it is later used as Base in the same file - the solution should keep the model intact and not destroy the already implemented logic, the base should be exported as it was - successfully.
- If Model is imported first and only later the Base is imported to a different Dataset (object in Catalog), we should try to link them up
    - We use the latest released (`STABLE`) version of the model that is referenced via the row where `Base` is defined.
    - Error is thrown in case there are no models found that could be referenced.

#### Notes:
- Covered extensively with test cases.

### DIFF
--- a/tests/structure/test_services.py
+++ b/tests/structure/test_services.py
@@ -2669,3 +2669,150 @@ def test_structure_with_origin_source_type_headers(app: DjangoTestApp):
         "datasets/gov/ivpk/adp",
     ]
     assert Comment.objects.filter(type=Comment.STRUCTURE_ERROR).count() == 0
+
+
+class TestStructureBaseModels:
+    @pytest.mark.django_db
+    def test_structure_base_and_model_defined_in_file(self, app: DjangoTestApp):
+        """Model for Base is defined in the same file."""
+        manifest = (
+            "id,dataset,resource,base,model,property,type,ref,source,prepare,level,status,visibility,access,uri,eli,title,description\n"
+            ",example,,,,,,,,,,,,,,,,\n"
+            ",,,,Animal,,,,,,0,completed,public,,,,,\n"
+            ",,,,,id,string,,source_animal_id,,4,completed,package,protected,,,,\n"
+            ",,,Animal,,,,,,,1,completed,public,,,,,\n"
+            ",,,,Dog,,,,,,0,completed,public,,,,,\n"
+            ",,,,,action,string,,source_dog_action,,4,completed,package,protected,,,,\n"
+            ",,,/,,,,,,,,,,,,,,\n"
+        )
+
+        structure = DatasetStructureFactory(file=FilerFileFactory(file=FileField(filename="file.csv", data=manifest)))
+        structure.dataset.current_structure = structure
+        structure.dataset.save()
+        create_structure_objects(structure)
+
+        assert Base.objects.count() == 1
+        base_object = Base.objects.get(metadata__name="example/Animal")
+        assert Base.objects.first().model.name == "Animal"
+        assert Model.objects.count() == 2
+        assert Model.objects.get(metadata__name="example/Animal")
+        assert Model.objects.get(metadata__name="example/Dog").base == base_object
+
+    @pytest.mark.django_db
+    def test_structure_two_imports_first_model_secondly_reference_the_model_as_base(self, app: DjangoTestApp):
+        """Two imports: First one imports a manifest with a model. Second, uses the model as a base.
+
+        In this case, the model that was imported with the first file must have a Version status of STABLE.
+        """
+        manifest_with_model_definition = (
+            "id,dataset,resource,base,model,property,type,ref,source,prepare,level,status,visibility,access,uri,eli,title,description\n"
+            ",example,,,,,,,,,,,,,,,,\n"
+            ",,,,Animal,,,,,,0,completed,public,,,,,\n"
+            ",,,,,id,string,,source_animal_id,,4,completed,package,protected,,,,\n"
+        )
+        structure_model = DatasetStructureFactory(
+            file=FilerFileFactory(file=FileField(filename="file.csv", data=manifest_with_model_definition))
+        )
+        structure_model.dataset.current_structure = structure_model
+        structure_model.dataset.save()
+        version = create_structure_objects(structure_model)
+        version.status = VersionStatus.STABLE
+        version.save()
+
+        manifest_with_base_reference = (
+            "id,dataset,resource,base,model,property,type,ref,source,prepare,level,status,visibility,access,uri,eli,title,description\n"
+            ",example2,,,,,,,,,,,,,,,,\n"
+            ",,,example/Animal,,,,,,,1,completed,public,,,,,\n"
+            ",,,,Dog,,,,,,0,completed,public,,,,,\n"
+            ",,,,,action,string,,source_dog_action,,4,completed,package,protected,,,,\n"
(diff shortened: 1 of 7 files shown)


tests/structure/test_services.py
```diff
@@ -2669,3 +2669,150 @@ def test_structure_with_origin_source_type_headers(app: DjangoTestApp):
         "datasets/gov/ivpk/adp",
     ]
     assert Comment.objects.filter(type=Comment.STRUCTURE_ERROR).count() == 0
+
+
+class TestStructureBaseModels:
+    @pytest.mark.django_db
+    def test_structure_base_and_model_defined_infile(self, app: DjangoTestApp):
```
martynasracct:
```suggestion
    def test_structure_base_and_model_defined_in_file(self, app: DjangoTestApp):
```

tests/structure/test_views.py
```diff
@@ -6894,6 +6901,148 @@ def test_export__duplicated_source_exports_resources_correctly(self, app: Django
             ",,,,,,,,,,,,,,,,,,,,",
         ]
 
+    def test_export__model_and_reference_as_base_infile(self, app: DjangoTestApp):
```
martynasracct:
```suggestion
    def test_export__model_and_reference_as_base_in_file(self, app: DjangoTestApp):
```

martynasracct (approved):
Good job! Feel free to ignore these comments

vitrina/structure/services.py
```diff
@@ -745,30 +745,46 @@ def _link_models(dataset: Dataset, dataset_meta: struct.Dataset, metadata_versio
...
 
 
+def resolve_base_model(base_metadata: struct.Base, metadata_version: Version) -> Model | None:
+    """Retrieves a base model based on metadata, prioritizing specific version or stable status.
+
+    This function attempts to find a Model object associated with a given base_metadata.
+    It handles two scenarios for model import:
+        1.  Model and Base in the same file: In this case, the function expects to find the model
+            with the same `metadata_version` as provided.
+        2.  Model imported before Base: If the model is defined in a separate file and then referenced
```
Gustascct:
```suggestion
        2.  Base imported before Model: If the model is defined in a separate file and then referenced
```

Shouldn't it be like this? The base model imported before the referencing model

tests/structure/test_services.py
Struchid: Adjusted and ammended the commit, thanks!:) [09736e3](https://github.com/atviriduomenys/katalogas/pull/2445/commits/09736e33132af8d48c959fbcdca9b833bb4fba37)

tests/structure/test_views.py
Struchid: Adjusted and ammended the commit, thanks!:) [09736e3](https://github.com/atviriduomenys/katalogas/pull/2445/commits/09736e33132af8d48c959fbcdca9b833bb4fba37)

Gustascct (approved):

vitrina/locale/en/LC_MESSAGES/django.po
```diff
@@ -2978,6 +2978,9 @@ msgstr "Delete scope"
 msgid "Organizacijų sujungimas"
 msgstr "Organization merge"
 
+msgid "Organiacijų sujungimas"
```
MindaugasStanionis:
```suggestion
msgid "Organizacijų sujungimas"
```

MindaugasStanionis (approved):

vitrina/locale/en/LC_MESSAGES/django.po
Struchid: Not mine, but fixed :p [5b7c708](https://github.com/atviriduomenys/katalogas/pull/2445/commits/5b7c7080df92221f61770d3c39b32d52ccff9c70)

vitrina/structure/services.py
Struchid: Yup, thanks! [5b7c708](https://github.com/atviriduomenys/katalogas/pull/2445/commits/5b7c7080df92221f61770d3c39b32d52ccff9c70)

PR #2477, opened 2026-03-10
[No task] Fix flaky test occasionally breaking CI/CD

Struchid:
CI/CD runs sometimes fail, due to a flaky test. Fixing the test to ensure that the run is smooth and without flaky exceptions.

Ensuring that the test checks what it needs to check and not some other objects that do not matter to the test.


Gustascct (approved):

MindaugasStanionis (approved):
🔥

PR #2483, opened 2026-03-17
2255: Organization as Representative

### LINKED ISSUE #2255: Galimybė priskirti organizaciją kaip tvarkytoją
**Specifikacijos punktas:**
- https://github.com/atviriduomenys/katalogas/issues/1427
#

VDA buvo priskirta kaip tvarkytoja apie tarkim 450 institucijų ir galėjo kurti bei tvarkyti duomenų rinkinius, dabar jie to nebegali daryti. Aiškinuosi ar jie turi pasirinkimus tvarkytojo redagavimo lange "Atvirų duomenų tvarkytojas" ir "Informacinės sistemos tvarkytojas", nes aš kaip super useris VDA nei pas viena tvarkytoją nematau tokių pasirinkimų/varnelių. **YRA WORKAROUND** (Aurimas tiksliau žino, bet iš esmės man atrodo reikėjo pridėti nauja VDA emailą kaip tvarkytoja prie visu organizacijų

MindaugasStanionis:
Summary:
- Add organization representative access via org chain
This PR extends the permission system to support indirect access — users can now gain roles through their organization being assigned as a representative, rather than only through direct assignment.
Core changes:
  - Added get_effective_user_role_via_organization on the Dataset model — resolves a user's effective role by walking the organization representative chain. Takes the most restrictive role between the org's assignment on the resource and the user's role within that organization.
  - Updated has_perm to check a third access path: if a user is a representative of an organization that is itself assigned as a representative on the resource hierarchy, they get access accordingly. 
  - Updated_filter_datasets_for_user in PermittedDatasetManager to include datasets accessible via the organization representative chain.
  - Adding check to not let open_data_representatives to create/update structure objects with visibility lower then 'Package'

Ticket: https://github.com/atviriduomenys/katalogas/issues/2255


vitrina/users/models.py
```diff
@@ -190,12 +196,35 @@ def is_resource_coordinator_for(self, organization: Organization | None) -> bool
...
+        open_data_roles = [Representative.OPEN_DATA_COORDINATOR, Representative.OPEN_DATA_MANAGER]
+
+        # Case 1: user is directly a representative of the organization
+        if Representative.objects.filter(
             user=self,
-            content_type=org_type,
+            content_type=org_content_type,
+            object_id=organization.pk,
+            role__in=open_data_roles,
+        ).exists():
```
martynasracct: Optional:
maybe it would be a bit clearer if "base" representative query is moved to variable:

```suggestion
        open_organization_representative_queryset = Representative.objects.filter(
            content_type=org_content_type,
            object_id=organization.pk,
            role__in=open_data_roles,
        )
        if open_organization_representative_queryset.filter(user=self).exists():
```

And then at the end of function:
```python
        return open_organization_representative_queryset.filter(
            role__in=open_data_roles, organization__in=user_org_ids,
        ).exists()
```

vitrina/datasets/services.py
```diff
@@ -304,6 +314,27 @@ def filter_out_non_public_datasets_for_user(user: User, datasets: SearchQuerySet
...
+    if user_org_ids:
+        resource_roles = (Representative.RESOURCE_MANAGER, Representative.RESOURCE_COORDINATOR)
+        open_data_roles = (Representative.OPEN_DATA_MANAGER, Representative.OPEN_DATA_COORDINATOR)
+
+        org_chain_qs = Representative.objects.filter(
+            organization__in=user_org_ids,
+            content_type__in=[org_ct, dataset_ct],
+        ).values_list("content_type_id", "role", "object_id")
+
+        for ct_id, role, obj_id in org_chain_qs:
```
martynasracct: More descriptive names
```suggestion
        # content_type_id and object_id - can be either organization or dataset
        for content_type_id, role, object_id in organization_chain_queryset:
```

vitrina/datasets/models.py
```diff
@@ -855,8 +855,45 @@ def get_acl_parents(self):
     def get_members_url(self):
         return reverse("dataset-members", kwargs={"pk": self.pk})
 
+    def get_effective_user_role_via_organization(self, user: "User") -> str | None:
```
martynasracct, line 858: Optional: I would love a small comment about what this method does. Mainly `max()` and `min()` parts.

tests/datasets/test_managers.py
```diff
@@ -214,3 +214,243 @@ def test_view_permissions_resource_managers(
...
+            ("regular_user", "grandchild", True, "CONFIDENTIAL", "dataset", False),
+            ("random_org_representative", "grandchild", True, "CONFIDENTIAL", "dataset", False),
+            ("org_representative", "grandchild", True, "CONFIDENTIAL", "dataset", False),
+            ("data_set_representative", "grandchild", True, "CONFIDENTIAL", "dataset", False),
+            ("parent_representative", "grandchild", True, "CONFIDENTIAL", "dataset", False),
+            ("global_representative", "grandchild", True, "CONFIDENTIAL", "dataset", True),
+        ],
+    )
+    def test_view_permissions_open_data_managers(
+        self, user_attributes, dataset_attributes, is_public, access_rights, subclass, expected
```
martynasracct: Missing type hints

tests/datasets/test_managers.py
```diff
@@ -214,3 +214,243 @@ def test_view_permissions_resource_managers(
...
+            ("regular_user", "grandchild", True, "CONFIDENTIAL", "dataset", False),
+            ("random_org_representative", "grandchild", True, "CONFIDENTIAL", "dataset", False),
+            ("org_representative", "grandchild", True, "CONFIDENTIAL", "dataset", True),
+            ("data_set_representative", "grandchild", True, "CONFIDENTIAL", "dataset", True),
+            ("parent_representative", "grandchild", True, "CONFIDENTIAL", "dataset", True),
+            ("global_representative", "grandchild", True, "CONFIDENTIAL", "dataset", True),
+        ],
+    )
+    def test_view_permissions_resource_managers(
+        self, user_attributes, dataset_attributes, is_public, access_rights, subclass, expected
```
martynasracct: missing type hints

tests/orgs/test_services.py
```diff
@@ -196,6 +231,67 @@ def test_dataset_permissions(role, action, dataset_fixture, expected, request):
...
+        ("open_data_manager", Action.UPDATE, "confidential_dataset", False),
+        ("open_data_manager", Action.VIEW, "public_dataset", True),
+        ("open_data_manager", Action.VIEW, "non_public_dataset", False),
+        ("open_data_manager", Action.VIEW, "confidential_dataset", False),
+        ("open_data_manager", Action.DELETE, "public_dataset", True),
+        ("open_data_manager", Action.DELETE, "non_public_dataset", False),
+        ("open_data_manager", Action.DELETE, "confidential_dataset", False),
+    ],
+)
+def test_dataset_permissions_via_org_representative(role, action, dataset_fixture, expected, request):
```
martynasracct: type hints

tests/orgs/test_services.py
```diff
@@ -505,19 +795,107 @@ def test_organization_representative_view_permission_manager(role: str, expected
...
+    RepresentativeFactory(
+        content_type=rep_org_ct,
+        object_id=representative_org.pk,
+        role=role,
+        user=user,
+        organization=None,
+    )
+
+    res = has_perm(user, Action.UPDATE, Representative, organization)
+    assert res is expected
```
martynasracct:
```suggestion
    assert has_perm(user, Action.UPDATE, Representative, organization) is expected
```

martynasracct (approved):
LGTM. Only small nitpickish comments 🙇

wisecrow (approved):

tests/orgs/test_services.py
MindaugasStanionis: Fixed in 1d159bf1da2b2731587b65372a0c283738a6e05d

tests/datasets/test_managers.py
MindaugasStanionis: Fixed in 1d159bf1da2b2731587b65372a0c283738a6e05d

tests/datasets/test_managers.py
MindaugasStanionis: Fixed in 1d159bf1da2b2731587b65372a0c283738a6e05d

vitrina/datasets/models.py
MindaugasStanionis: Fixed in 1d159bf1da2b2731587b65372a0c283738a6e05d

vitrina/datasets/services.py
MindaugasStanionis: Fixed in 1d159bf1da2b2731587b65372a0c283738a6e05d

vitrina/users/models.py
MindaugasStanionis: Fixed in 1d159bf1da2b2731587b65372a0c283738a6e05d

Gustascct (approved):

vitrina/users/models.py
```diff
@@ -190,12 +196,33 @@ def is_resource_coordinator_for(self, organization: Organization | None) -> bool
     def is_open_data_representative_for(self, organization: Organization | None) -> bool:
         if organization is None:
             return False
-        org_type = self.organization_content_type
-        return Representative.objects.filter(
-            user=self,
-            content_type=org_type,
+
+        org_content_type = self.organization_content_type
```
Struchid:
```suggestion
        organization_content_type = self.organization_content_type
```

vitrina/users/models.py
```diff
@@ -190,12 +196,33 @@ def is_resource_coordinator_for(self, organization: Organization | None) -> bool
...
+        open_data_roles = [Representative.OPEN_DATA_COORDINATOR, Representative.OPEN_DATA_MANAGER]
+
+        # Case 1: user is directly a representative of the organization
+        open_organization_representative_queryset = Representative.objects.filter(
+            content_type=org_content_type,
             object_id=organization.pk,
-            role__in=[Representative.OPEN_DATA_COORDINATOR, Representative.OPEN_DATA_MANAGER],
+            role__in=open_data_roles,
+        )
+        if open_organization_representative_queryset.filter(user=self).exists():
```
Struchid: Shouldn't we also add the `deleted=False` filter?

vitrina/requests/views.py
```diff
@@ -764,46 +764,41 @@ def get_context_data(self, **kwargs):
...
+                and Dataset.objects.filter(pk__in=dataset_ids)
+                .exclude(access_rights__in=[Dataset.PUBLIC, Dataset.RESTRICTED])
+                .exists()
+            ):
+                return False
+
+        return any(
+            has_perm(user, Action.CREATE, RequestAssignment, request_assignment.organization)
+            for request_assignment in self.request_obj.requestassignment_set.all()
+        )
```
Struchid: Two nitpicky suggestions:

- `if has_open_data_role` can be reversed (`if not has_open_data_role`) for an early return and less indented code
- For the inner checks:
```
if (
                dataset_ids
                and Dataset.objects.filter(pk__in=dataset_ids)
                .exclude(access_rights__in=[Dataset.PUBLIC, Dataset.RESTRICTED])
                .exists()
            ):
```

So:

```
filtered_datasets = Dataset.objects.filter(pk__in=dataset_ids).exclude(access_rights__in[Dataset.PUBLIC, Dataset.RESTRICTED]).exists()
if (
                dataset_ids
                and filtered_datasets
            ):
```

Let's move the second queryset to a variable as well, will definitely increase readability

vitrina/requests/views.py
```diff
@@ -1377,20 +1372,29 @@ def has_permission(self):
...
+            role__in=Representative.OPEN_DATA_ROLE_KEYS,
+        ).exists()
+
+        dataset_filter = Q(organization=user.organization)
+        if is_open_data_role:
+            dataset_filter &= Q(access_rights__in=(Dataset.PUBLIC, Dataset.RESTRICTED))
+
+        context_data["form"].fields["datasets"].queryset = Dataset.objects.filter(dataset_filter)[
+            : self.dataset_query_limit
+        ]
```
Struchid: Are the users able to reach the other datasets? Or is this a hard limit and will remove any datasets that are behind that limit?

vitrina/orgs/models.py
```diff
@@ -186,6 +186,8 @@ class Representative(models.Model):
     OPEN_DATA_ROLE_KEYS = {role[0] for role in OPEN_DATA_ROLES}
     RESOURCE_ROLE_KEYS = {role[0] for role in RESOURCE_ROLES}
 
+    ROLE_HIERARCHY = [role for role in MANAGER_ROLES]
```
Struchid: Why would you need this? This is the same as `MANAGER_ROLES` but as a list. If you need the information, that they are ordered hierarchically, a comment could be left, other than that, why not just use `MANAGER_ROLES` for wherever you are using `ROLE_HIERARCHY`?

Also, if you really do need this, this does the same:

```suggestion
    ROLE_HIERARCHY = list(MANAGER_ROLES)
```

vitrina/orgs/services.py
```diff
@@ -478,29 +478,10 @@ def determine_user_role(user: User, resource: Dataset) -> Role:
...
-            ]
-        )
-        .filter(user=user)
-        .exists()
-    ):
-        return Role.OPEN_DATA_COORDINATOR if user.is_open_data_coordinator else Role.OPEN_DATA_MANAGER
+    if role := user.get_representative_role_for_resource(resource):
+        return Role(role)
+    if role := resource.get_effective_user_role_via_organization(user):
+        return Role(role)
```
Struchid: **Optional:** Can be combined, but decide yourself based on what's more readable to you (yours or my suggestion):


```suggestion
    if role := (
        user.get_representative_role_for_resource(resource) or
        resource.get_effective_user_role_via_organization(user)
    )
        return Role(role)
```

vitrina/orgs/services.py
```diff
@@ -622,6 +603,24 @@ def has_perm(
...
+                user_representative_organizations = Representative.objects.filter(
+                    user=user,
+                    content_type=representative_organization_ct,
+                    object_id__in=representative_organization_ids,
+                )
+
+                if user_representative_organizations.exists():
+                    if isinstance(obj, Representative):
+                        return obj.can_be_updated_by(user)
+                    return True
```
Struchid: **Nitpick:**


```suggestion
                    return obj.can_be_updated_by(user) if isinstance(obj, Representative) else True
```

vitrina/orgs/forms.py
```diff
@@ -456,12 +456,19 @@ def __init__(self, *args, **kwargs):
...
 
-        self.fields["role"].choices = allowed_roles
+        content_type = ContentType.objects.get_for_model(self.object_model)
+
+        is_open_data_coordinator = Representative.objects.filter(
+            user=self.user,
+            content_type=content_type,
+            object_id=self.object.pk,
+            role=Representative.OPEN_DATA_COORDINATOR,
+        ).exists()
```
Struchid: I guess you can already make some function for this on the `Representative` model for reuse

vitrina/datasets/managers.py
```diff
@@ -53,18 +53,44 @@ def _filter_datasets_for_user(self, user: "User", datasets: QuerySet["Dataset"])
...
+            organization__in=user_org_ids,
+        ).values_list("object_id", flat=True)
+
         represented_paths = set(
             Dataset.objects.filter(
-                Q(pk__in=represented_dataset_ids) | Q(organization_id__in=represented_org_ids)
+                Q(pk__in=represented_dataset_ids)
+                | Q(organization_id__in=represented_org_ids)
+                | Q(pk__in=org_represented_dataset_ids)
+                | Q(organization_id__in=org_represented_org_ids)
```
Struchid: Could be combined:


```suggestion
        Q(pk__in=represented_dataset_ids + org_represented_dataset_ids)
        | Q(organization_id__in=represented_org_ids + org_represented_org_ids)
)
```

vitrina/datasets/models.py
```diff
@@ -855,8 +855,55 @@ def get_acl_parents(self):
...
+            Q(content_type=dataset_ct, object_id__in=dataset_ids)
+            | Q(content_type=organization_ct, object_id__in=organization_ids),
+            organization_id__in=user_org_map.keys(),
+        ).values_list("organization_id", "role")
+
+        effective_roles = []
+        for org_id, org_role in org_roles:
+            user_role = user_org_map[org_id]
+            # Take the weaker of the org's role and the user's role in that org.
+            # e.g. org has "resource_manager", user is "open_data_manager" in that org → effective role is "open_data_manager"
```
Struchid:
```suggestion
            # Use the more restrictive role between the organization's role and the user's role within that org.
            # e.g. org has RESOURCE_MANAGER, user has OPEN_DATA_MANAGER → effective role is OPEN_DATA_MANAGER
```

vitrina/datasets/services.py
```diff
@@ -273,9 +273,19 @@ def filter_out_non_public_datasets_for_user(user: User, datasets: SearchQuerySet
...
 
+    # Collect organizations the user belongs to via Representative records and user.organization
+    user_org_ids = list(
+        Representative.objects.filter(
+            user=user,
+            content_type=organization_ct,
+        ).values_list("object_id", flat=True)
+    )
+    if user.organization:
+        user_org_ids.append(user.organization.pk)
```
Struchid: FK value should be reachable directly:

```suggestion
        user_org_ids.append(user.organization_id)
```

- https://docs.djangoproject.com/en/4.0/topics/db/optimization/#use-foreign-key-values-directly

Struchid (approved):

vitrina/requests/views.py
MindaugasStanionis: The limit only applies to the initially loaded results. Users can still find any dataset by typing in the Select2 search field.

vitrina/datasets/managers.py
MindaugasStanionis: QuerySets don't support the + operator for concatenation, so I will leave as is

vitrina/datasets/services.py
MindaugasStanionis: Thanks, good point

vitrina/users/models.py
```diff
@@ -1,4 +1,7 @@
+from __future__ import annotations
```
martynasracct: I don't think it's needed. Katalogas does not support python versions without typing

vitrina/structure/forms.py
```diff
@@ -617,8 +641,9 @@ class Meta:
             "is_parameterized",
         )
 
-    def __init__(self, dataset, metadata_version, *args, **kwargs):
+    def __init__(self, request: WSGIRequest, dataset: Dataset, metadata_version: Version, *args, **kwargs):
```
martynasracct: As I can see, request is only used for `request.user`. If that's the case - we should send user object to this form. Also, it is not used anywhere outside `__init__` method, so there is no point assigning it to `self.request`

vitrina/structure/forms.py
```diff
@@ -163,9 +165,10 @@ class Meta:
             "description",
         )
 
-    def __init__(self, prop: Property, enum: Enum, *args, **kwargs):
+    def __init__(self, request: WSGIRequest, prop: Property, enum: Enum, *args, **kwargs):
```
martynasracct: As I can see, request is only used for `request.user`. If that's the case - we should send user object to this form. Also, it is not used anywhere outside `__init__` method, so there is no point assigning it to `self.request`

vitrina/requests/views.py
```diff
@@ -1377,20 +1377,29 @@ def has_permission(self):
...
+            content_type=ContentType.objects.get_for_model(Organization),
+            object_id=user.organization.pk,
+            role__in=Representative.OPEN_DATA_ROLE_KEYS,
+        ).exists()
+
+        dataset_filter = Q(organization=user.organization)
+        if is_open_data_role:
+            dataset_filter &= Q(access_rights__in=(Dataset.PUBLIC, Dataset.RESTRICTED))
+
+        context_data["form"].fields["datasets"].queryset = Dataset.objects.filter(dataset_filter)[
```
martynasracct: Can we move this to form?

vitrina/requests/views.py
```diff
@@ -1402,16 +1411,23 @@ class RequestDatasetsEditUpdateView(RequestDatasetsEditView):
...
+
+        queryset = Dataset.objects.filter(organization=user.organization)
+        if is_open_data_role:
             queryset = queryset.filter(access_rights__in=[Dataset.PUBLIC, Dataset.RESTRICTED])
         if term:
             queryset = queryset.filter(translations__title__istartswith=term)
-        queryset = queryset.order_by("translations__title")[: self.queryset_limit]
-        form.fields["datasets"].queryset = queryset
+
+        form.fields["datasets"].queryset = queryset.order_by("translations__title")[: self.queryset_limit]
```
martynasracct: Can we move this to form?

vitrina/orgs/forms.py
```diff
@@ -494,13 +496,11 @@ def clean(self):
...
-                self.add_error(
-                    "role",
-                    _("Jūs neturite teisės priskirti šios rolės."),
-                )
+        if role:
+            if Representative.is_open_data_coordinator(self.user, content_type, self.object.pk):
+                allowed_roles = dict(Representative.OPEN_DATA_ROLES)
+                if role not in allowed_roles:
+                    self.add_error("role", _("Jūs neturite teisės priskirti šios rolės."))
         return super().clean()
```
martynasracct:
```suggestion
        if role and role not in dict(Representative.OPEN_DATA_ROLES) and Representative.is_open_data_coordinator(self.user, content_type, self.object.pk):
            self.add_error("role", _("Jūs neturite teisės priskirti šios rolės."))
return super().clean()
```

martynasracct (approved):

tests/datasets/test_managers.py
```diff
@@ -214,3 +214,355 @@ def test_view_permissions_resource_managers(
...
+
+@pytest.mark.django_db
+class TestDatasetViewPermissionsViaOrgRepresentative:
+    def setup_method(self):
+        self.main_organization = OrganizationFactory(kind=Organization.COM)
+        self.random_organization = OrganizationFactory(kind=Organization.GOV)
+        self.grand_organization = OrganizationFactory(kind=Organization.COM)
+
+        # Create dataset hierarchy
+        self.grand_parent = DatasetFactory(is_public=True, organization=self.grand_organization)
```
Struchid: **Nitpick:** Grandparent is a single word (just like grandchild)

vitrina/datasets/forms.py
```diff
@@ -225,6 +247,13 @@ def __init__(
 
             self.initial["applicable_legislation"] = list(instance.applicable_legislation.values_list("url", flat=True))
             self.initial["documentation"] = list(instance.documentation.values_list("documentation_link", flat=True))
+
+        if is_open_data_representative:
+            self.fields["access_rights"].choices = [
+                (Dataset.PUBLIC, _("Vieši")),
+                (Dataset.RESTRICTED, _("Apriboti")),
+            ]
```
Struchid: This is duplicated with lines 237-241, no?

vitrina/structure/views.py
```diff
@@ -2033,6 +2034,7 @@ def has_permission(self):
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
         kwargs["prop"] = self.property
+        kwargs["user"] = self.request.user
```
Struchid, line 2037: **Small and optional suggestion:** What about a base class that defines this for Enum classes? I see it being reused in multiple classes

vitrina/datasets/models.py
```diff
@@ -855,8 +855,58 @@ def get_acl_parents(self):
...
+            # Use the more restrictive role between the organization's role and the user's role within that org.
+            # e.g. org has RESOURCE_MANAGER, user has OPEN_DATA_MANAGER → effective role is OPEN_DATA_MANAGER
+            effective_roles.append(max(org_role, user_role, key=Representative.MANAGER_ROLES.index))
+
+        if not effective_roles:
+            return None
+
+        # If the user represents multiple orgs, return the most privileged role across all of them.
+        # e.g. ["open_data_manager", "resource_manager"] → "resource_manager"
+        return min(effective_roles, key=Representative.MANAGER_ROLES.index)
```
Struchid, line 906: Just double checking but the most `min` role is the most privileged one, right? (Following the information in the comment and in the logic).

Struchid (approved):

vitrina/datasets/models.py
MindaugasStanionis: Yes, Manager_Roles is: 
MANAGER_ROLES = (
        RESOURCE_MANAGER,
        OPEN_DATA_MANAGER,
    )

vitrina/users/models.py
```diff
@@ -91,14 +95,16 @@ def get_acl_parents(self):
     def organization_content_type(self):
         return ContentType.objects.get_for_model(Organization)
 
-    def get_representative_role(self) -> str:
-        if self.is_resource_coordinator:
+    def get_representative_role_for_resource(self, resource: "Dataset") -> str:
+        if self.pk in resource.get_managers_queryset([Representative.RESOURCE_COORDINATOR]):
```
wisecrow:
```suggestion
        if resource.get_managers_queryset([Representative.RESOURCE_COORDINATOR]).filter(user=self).exists():
```

wisecrow (approved):

vitrina/requests/views.py
MindaugasStanionis: I will leave it in the view, because form does not have a user, and in my opinion, this thing is an authorization/access concern — it should belong to the view. Or you think it's better to move this to form>

martynasracct: I wouldn't be a first from in Katalogas, that uses user model.
Another way would be to override view method `.get_form(..)`. But I would move it to form, to have all form logic in form class

MindaugasStanionis: Done that, thanks for suggestion

vitrina/structure/views.py
MindaugasStanionis: Will leave this for now, don't have much time in this task

PR #2505, opened 2026-03-26
Fix projects permission

kastger:
image
image

### DIFF
--- a/vitrina/datasets/templates/vitrina/datasets/project_list.html
+++ b/vitrina/datasets/templates/vitrina/datasets/project_list.html
@@ -47,6 +47,11 @@
                                         <a href="{{ project.get_absolute_url }}" class="dataset-list-title">
                                               {{ project.get_title }}
                                         </a>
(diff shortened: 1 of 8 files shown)


martynasracct (approved):

Struchid (approved):
Side question: Is there a way to approve projects without going through DjAdmin?

kastger:
> Side question: Is there a way to approve projects without going through DjAdmin?

I don't think there is, I've raised this question before (months ago) but there was more high priority changes. I will try to raise the question again and create a task so it doesn't get forgotten again.

Struchid:
> > Side question: Is there a way to approve projects without going through DjAdmin?
> 
> I don't think there is, I've raised this question before (months ago) but there was more high priority changes. I will try to raise the question again and create a task so it doesn't get forgotten again.

I do not think it's an issue as of right now (while these are not used very frequently), just wanted to know if it was possible due to a task I have regarding these use cases :D

gts446:
I think you can approve through here 
image

PR #2528, opened 2026-04-03
[538]: Updating is_open_data_representative_for to check representative also for ancestor datasets

MindaugasStanionis:
Summary:

- Updated `is_open_data_representative_for` function to add a optional parameter `role` to check coordinators explicitly.
- Updated `is_open_data_representative_for` function to check for objects which has get_ancestors functionality.

Ticket: https://github.com/atviriduomenys/dvms/issues/538

### DIFF
--- a/vitrina/users/models.py
+++ b/vitrina/users/models.py
@@ -210,16 +210,13 @@ class User(AbstractUser):
         manager_equivalent_roles = [coordinator_to_manager.get(role, role) for role in open_data_roles]
 
         # Case 1: user directly holds an open data role on the object
-        if (
-            Representative.objects.filter(
-                content_type=content_type,
-                object_id=obj.pk,
-                role__in=open_data_roles,
-                user=self,
-            )
-            .exclude(deleted=True)
-            .exists()
-        ):
+        direct_representative = Representative.objects.filter(
+            content_type=content_type,
+            object_id=obj.pk,
+            role__in=open_data_roles,
+            user=self,
+        ).exclude(deleted=True)
+        if direct_representative.exists():
             return True
 
         # Case 2: access is resolved through the user's organizational membership.
@@ -234,7 +231,7 @@ class User(AbstractUser):
         )
 
         if user_organization_role_map:
-            for organization_id, organization_role in (
+            organization_representatives = (
                 Representative.objects.filter(
                     content_type=content_type,
                     object_id=obj.pk,
@@ -242,7 +239,9 @@ class User(AbstractUser):
                 )
                 .exclude(deleted=True)
                 .values_list("organization_id", "role")
-            ):
+            )
+
+            for organization_id, organization_role in organization_representatives:
                 user_role = coordinator_to_manager.get(
                     user_organization_role_map[organization_id], user_organization_role_map[organization_id]
                 )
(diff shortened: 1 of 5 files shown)


vitrina/users/models.py
```diff
@@ -226,27 +233,34 @@ def is_open_data_representative_for(self, obj: Union[Organization, "Dataset", No
...
+        # Case 3: check ancestors if the object supports it
+        if hasattr(obj, "get_ancestors"):
+            for ancestor in obj.get_ancestors():
+                if self.is_open_data_representative_for(ancestor, roles=roles):
+                    return True
 
         return False
 
+    def is_open_data_coordinator_for(self, obj: Union[Organization, "Dataset", None]) -> bool:
+        return self.is_open_data_representative_for(obj, roles=[Representative.OPEN_DATA_COORDINATOR])
```
Struchid, line 268: The method name says `_coordinator_`, but the function called in the method states `representative` is this an expected naming clash?

vitrina/users/models.py
```diff
@@ -194,13 +194,20 @@ def is_resource_coordinator_for(self, organization: Organization | None) -> bool
...
+        obj: Union[Organization, "Dataset", None],
+        roles: list | None = None,
+    ) -> bool:
         if obj is None:
             return False
 
         content_type = ContentType.objects.get_for_model(obj)
-        open_data_roles = [Representative.OPEN_DATA_COORDINATOR, Representative.OPEN_DATA_MANAGER]
+        open_data_roles = (
+            roles if roles is not None else [Representative.OPEN_DATA_COORDINATOR, Representative.OPEN_DATA_MANAGER]
```
Struchid, line 207: **Double-checking:** What if `roles` is an empty list? Do we still want the `open_data_roles` to be an empty list or should we set the default roles for that case as well?

vitrina/users/models.py
```diff
@@ -226,27 +233,34 @@ def is_open_data_representative_for(self, obj: Union[Organization, "Dataset", No
...
+        if user_organization_role_map:
+            for organization_id, organization_role in (
+                Representative.objects.filter(
+                    content_type=content_type,
+                    object_id=obj.pk,
+                    organization_id__in=user_organization_role_map.keys(),
+                )
+                .exclude(deleted=True)
+                .values_list("organization_id", "role")
+            ):
```
Struchid: **Optional:** I'd extract this queryset into a separate variable, takes quite a lot of lines and gets harder to read (personal opinion)

Struchid (approved):

vitrina/users/models.py
MindaugasStanionis: The naming is intentional — `coordinator` is a specific role within the broader `representative` class, so the method just delegates to the general checker with that role hardcoded. I'll add a docstring to avoid this confusion in the future.

vitrina/users/models.py
MindaugasStanionis: Calling this function with empty list would be a callers mistake, not a intentional bug. This case is only for is_open_data_coordinator_for function

vitrina/users/models.py
Struchid: Nice! Thanks for the explanation :-)

PR #2531, opened 2026-04-07
[538]: Open Data Coordinator is not checked in ancestor datasets

MindaugasStanionis:
Summary:

- Updated is_open_data_representative_for function to add a optional parameter role to check coordinators explicitly.
- Updated is_open_data_representative_for function to check for objects which has get_ancestors functionality.

Ticket: https://github.com/atviriduomenys/dvms/issues/538

The branch has been renamed, because in drone the name was too long

### DIFF
--- a/vitrina/orgs/forms.py
+++ b/vitrina/orgs/forms.py
@@ -357,11 +357,17 @@ class RepresentativeUpdateForm(ModelForm):
         self.object = kwargs.pop("object", None)
         super().__init__(*args, **kwargs)
 
-        organization = self.object if self.object_model == Organization else getattr(self.object, "organization", None)
+        self.is_resource_coordinator = self.user.is_coordinator_for(
+            self.object, roles=[Representative.RESOURCE_COORDINATOR]
+        )
+        self.is_open_data_coordinator = self.user.is_coordinator_for(
+            self.object, roles=[Representative.OPEN_DATA_COORDINATOR]
+        )
+
+        # If they are JUST an OD coordinator, restrict them.
         self.fields["role"].choices = (
             Representative.OPEN_DATA_ROLES
-            if self.user.is_open_data_coordinator_for(self.object)
-            or (organization and self.user.is_open_data_coordinator_for(organization))
+            if self.is_open_data_coordinator and not self.is_resource_coordinator
             else Representative.ROLES
         )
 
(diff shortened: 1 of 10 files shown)


vitrina/orgs/forms.py
```diff
@@ -417,6 +425,15 @@ def clean(self):
...
             raise ValidationError(_("Organizacijai gali būti suteikta tik tvarkytojo rolė"))
 
+        organization = self.object if self.object_model == Organization else getattr(self.object, "organization", None)
+
+        is_allowed = self.user.is_open_data_coordinator_for(self.object) or (
+            organization and self.user.is_open_data_coordinator_for(organization)
+        )
+
+        if role and role not in dict(Representative.OPEN_DATA_ROLES) and is_allowed:
+            self.add_error("role", _("Jūs neturite teisės priskirti šios rolės."))
```
Struchid, line 439: Translations seem to be missing for this one. Judging from the fact that the .po file is not comitted, so might be translated already :d

Struchid (approved):

martynasracct (approved):

vitrina/users/models.py
```diff
@@ -194,55 +194,127 @@ def is_resource_coordinator_for(self, organization: Organization | None) -> bool
...
-        coordinator_to_manager = dict(zip(Representative.COORDINATOR_ROLES, Representative.MANAGER_ROLES))
+        if self.is_staff or self.is_superuser:
+            return True
+
+        parents = obj.get_acl_parents()
+
+        # Check every node for a direct role.
+        for parent in parents:
+            if self._has_direct_representative_role(parent, roles):
+                return True
```
wisecrow:
```suggestion
       return any(self._has_direct_representative_role(parent, roles) for parent in parents)
```

vitrina/users/models.py
```diff
@@ -194,55 +194,127 @@ def is_resource_coordinator_for(self, organization: Organization | None) -> bool
...
+
+        # Check every node for organizational membership.
+        for parent in parents:
+            if self._has_role_via_org_membership(parent, roles):
+                return True
```
wisecrow:
```suggestion
        return any(self._has_role_via_org_membership(parent, roles) for parent in parents)
```

vitrina/users/models.py
```diff
@@ -194,55 +194,127 @@ def is_resource_coordinator_for(self, organization: Organization | None) -> bool
...
+    def _has_role_via_org_membership(
+        self,
+        obj: Union[Organization, "Dataset", None],
+        roles: list[str] | None = None,
+    ) -> bool:
+        """
+        User belongs to an organization that holds a role on this object.
+        Effective permission = least privileged of (user's org role, org's role on object).
+        """
+        content_type = ContentType.objects.get_for_model(obj)
```
wisecrow, line 291: Just interesting, what we would get here if `object` is `None` ?
If result is `None`, should not we safe guard queries to filter not `None` ?

wisecrow (approved):

vitrina/users/models.py
MindaugasStanionis: This is called inside another function which safe guards and early returns if obj is none, but yes, better to safe guard this also

vitrina/users/models.py
MindaugasStanionis: if any(self._has_direct_representative_role(p, roles) for p in parents):
    return True 
We need to have it like this because it will always return true or false and we will never reach second check for org_membership

vitrina/users/models.py
MindaugasStanionis: Added it in c32a37d44c1a4877474fce7683b92633439eed10

vitrina/users/models.py
MindaugasStanionis: Added it in c32a37d44c1a4877474fce7683b92633439eed10 Thanks!

vitrina/orgs/forms.py
MindaugasStanionis: Yeah, it was translated already

PR #2547, opened 2026-04-14
[549]: Allowing coordinators to create / update organizations as repr

MindaugasStanionis:
Summary:

- Removed can_update_publishers permission check from HTML forms and the corresponding view
- Removed the early return that was blocking non-superusers
- Removed the superuser-only restriction when updating an organization

Ticket: https://github.com/atviriduomenys/dvms/issues/549

### DIFF
--- a/tests/orgs/test_views.py
+++ b/tests/orgs/test_views.py
@@ -390,7 +390,6 @@ def test_representative_create_organization_as_representative(app: DjangoTestApp
     ]
     form["email"] = organization1.email
     form["role"] = "resource_manager"
-    form.submit()
     response = form.submit()
     assert response.status_code == 200
     assert Representative.objects.filter(email="test_org1@test.com").count() == 1
@@ -411,7 +410,6 @@ def test_representative_create_organization_as_coordinator_role(
     ]
     form["email"] = organization1.email
     form["role"] = representative_role
-    form.submit()
     response = form.submit()
     assert response.status_code == 200
     assert "Organizacijai gali būti suteikta tik tvarkytojo rolė" in response.context["form"].errors["role"][0]
--- a/tests/orgs/test_views.py
+++ b/tests/orgs/test_views.py
@@ -391,8 +391,11 @@ def test_representative_create_organization_as_representative(app: DjangoTestApp
     form["email"] = organization1.email
     form["role"] = "resource_manager"
     response = form.submit()
-    assert response.status_code == 200
-    assert Representative.objects.filter(email="test_org1@test.com").count() == 1
+    assert response.status_code == 302
+    representative_qs = Representative.objects.filter(email="test_org1@test.com")
+    assert representative_qs.count() == 1
+    assert representative_qs.first().role == "resource_manager"
+    assert representative_qs.first().organization == organization1
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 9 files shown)


tests/orgs/test_views.py
```diff
@@ -380,6 +381,56 @@ def test_representative_create_without_user_for_two_organizations(app: DjangoTes
...
+    form = app.get(reverse("representative-create", kwargs={"pk": representative_data["organization"].pk})).forms[
+        "representative-form"
+    ]
+    form["email"] = organization1.email
+    form["role"] = representative_role
+    form.submit()
+    response = form.submit()
+    assert response.status_code == 200
+    assert "Organizacijai gali būti suteikta tik tvarkytojo rolė" in response.context["form"].errors["role"][0]
+    assert Representative.objects.filter(email="test_org1@test.com").count() == 0
```
Gustascct, line 419: Can forms send an error code instead of a 200?

tests/orgs/test_views.py
Gustascct, line 1: Are there any error paths tested for the updates/create? If not, could you add some?

Gustascct (approved):

vitrina/orgs/views.py
```diff
@@ -548,7 +548,6 @@ def get_context_data(self, **kwargs):
             Representative,
             self.organization,
         )
-        context_data["can_delete_publishers"] = self.request.user.is_superuser
```
Struchid, line 551: So without this, no one can delete publishers?

Struchid (approved):

vitrina/orgs/views.py
MindaugasStanionis: This was used to explicitly disallow coordinators from editing/deleting an organization as a representative if they are not superusers. Without it, all coordinators would be able to edit it based on their role, just like other representatives.

tests/orgs/test_views.py
MindaugasStanionis: I guess it could, but we need to override what status does form_invalid return

tests/orgs/test_views.py
MindaugasStanionis: Yes, error paths are covered — there are tests verifying that an open_data_coordinator cannot assign the resource_manager role, and that an organization as a representative cannot be assigned a coordinator role.

martynasracct (approved):

tests/orgs/test_views.py
```diff
@@ -380,6 +381,56 @@ def test_representative_create_without_user_for_two_organizations(app: DjangoTes
...
+    app.set_user(representative_data["resource_coordinator"])
+    organization1 = OrganizationFactory(email="test_org1@test.com")
+
+    form = app.get(reverse("representative-create", kwargs={"pk": representative_data["organization"].pk})).forms[
+        "representative-form"
+    ]
+    form["email"] = organization1.email
+    form["role"] = "resource_manager"
+    form.submit()
+    response = form.submit()
```
wisecrow, line 393: Do you need to call `form.submit()` twice?

tests/orgs/test_views.py
```diff
@@ -380,6 +381,56 @@ def test_representative_create_without_user_for_two_organizations(app: DjangoTes
...
+    app.set_user(representative_data[representative_role])
+    organization1 = OrganizationFactory(email="test_org1@test.com")
+
+    form = app.get(reverse("representative-create", kwargs={"pk": representative_data["organization"].pk})).forms[
+        "representative-form"
+    ]
+    form["email"] = organization1.email
+    form["role"] = representative_role
+    form.submit()
+    response = form.submit()
```
wisecrow, line 416: Calling `form.submit()` twice here also.

wisecrow (approved):

PR #2559, opened 2026-04-20
Add dcat app. Some code refactoring

martynasracct:
**Summary:**

* Adds `dcat` app. It will be used for new Dataset CRUD that only changes non-public datasets.
* Adds variable `Agency.RISR_CODE = "risr"` and uses it throughout the code.
* Moves some code from `Dataset` form and view to separate methods so that it can be reused. It's mostly copy-paste code:
  - `validate_dataset_name`
  - `validate_applicable_legislation`
  - `validate_identifier`
  - `create_tasks_and_notify_subscribers_about_dataset_creation`
  - `create_dataset_representative_and_attribution`


vitrina/datasets/models.py
```diff
@@ -1079,8 +1079,12 @@ def get_level(self) -> int | None:
 
     @property
     def identifier(self) -> str | None:
+        from vitrina.identifiers.models import Agency
+
```
Gustascct, line 1083: Is this solving a circular dependency? If not, please move it to the beginning of the file.

vitrina/datasets/form_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+    return name
+
+
+def validate_applicable_legislation(urls: list[str]) -> list[str | None]:
+    validator = URLValidator()
+    item_errors = []
+
+    for url in urls:
+        if not url:
+            item_errors.append(None)
```
Gustascct, line 92: I see that this is not yours, but why is it appending None if afterwards the item_errors is filtered and Nones are not used? Maybe we could remove it altogether?

Gustascct (approved):

vitrina/helpers.py
```diff
@@ -451,7 +451,7 @@ def submit(title=None):
     return (Submit("submit", title, css_class="button is-primary"),)
 
 
-def get_current_domain(request: WSGIRequest, ensure_secure=False) -> str:
+def get_current_domain(request: WSGIRequest | HttpRequest, ensure_secure=False) -> str:
```
Struchid: `WSGIRequest` inherits from `HttpRequest`, so you could leave the base type (`HttpRequest`) here IMO:


```suggestion
def get_current_domain(request: HttpRequest, ensure_secure=False) -> str:
```

vitrina/datasets/form_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+
+def validate_dataset_name(name: str | None, dataset: Dataset | None, organization: Organization) -> None:
+    existing_metadata = dataset.metadata.first() if dataset and dataset.pk else None
+
+    if name:
+        if not name.isascii():
+            raise ValidationError(_("Kodiniame pavadinime gali būti naudojamos tik lotyniškos raidės."))
+
+        if any(ch.isupper() for ch in name):
+            raise ValidationError(_("Kodiniame pavadinime gali būti naudojamos tik mažosios raidės."))
```
Struchid, line 25: These are old and translated strings and do not need to be translated, right?

eimantas-nejus-cct (approved):

vitrina/datasets/form_helpers.py
Struchid: It throws errors based on the index IIRC, that is why we need to prefill the `item_errors` with `None` if no error is thrown for the field (there could be an N amount of fields added for that field). If we would not do that, the error would get thrown for the first M amount of errors collected, but N > M, meaning that the errors thrown would be shifted to incorrect fields.

vitrina/datasets/form_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+
+def validate_identifier(identifier: str | None) -> None:
+    if not identifier:
+        return
+
+    agency = get_object_or_404(Agency, code=Agency.RISR_CODE)
+    is_regexp = agency.identifier_validation_type == Agency.IdentifierValidationType.REGEXP
+
+    if is_regexp and (pattern := agency.identifier_validation_options):
+        if not re.fullmatch(pattern, identifier):
```
Struchid: Seems like could be combined (leaving the walrus as is)

Struchid (approved):
Few optional comments from my side, otherwise LGTM 🧨

vitrina/datasets/models.py
martynasracct: Yes. Circular dependependency

vitrina/datasets/form_helpers.py
martynasracct: @Struchid is right. Errors are based on index.

vitrina/datasets/form_helpers.py
martynasracct: Yes, correct

vitrina/datasets/form_helpers.py
```diff
@@ -0,0 +1,119 @@
...
+        if metadata_qs.exists():
+            raise ValidationError(_("Duomenų rinkinys su šiuo kodiniu pavadinimu jau egzistuoja."))
+
+    else:
+        if existing_metadata and existing_metadata.name:
+            raise ValidationError(
+                _("Kodinis pavadinimas yra privalomas, jei duomenų rinkinys jau turi kodinį pavadinimą.")
+            )
+
+    return name
```
wisecrow: Fix annotation to -> `str | None` or drop the return.

wisecrow (approved):

MindaugasStanionis (approved):

vitrina/datasets/form_helpers.py
martynasracct: Good catch!

vitrina/datasets/form_helpers.py
martynasracct: 🦭

PR #2561, opened 2026-04-22
Move some more Dataset CRUD related code to separate methods

martynasracct:
**Summary:**
Moves some more code from Dataset form to separate methods so that it can be reused. It's mostly copy-paste code:
- `set_default_agent_endpoint_fields` - small refactoring to make code more readable
- `validate_agent_endpoint_fields` - small refactoring to make code more readable
- `get_contact_form_choices`


vitrina/dcat/urls.py
```diff
@@ -1 +1,16 @@
-urlpatterns = []
+from django.urls import path
+
+from vitrina.dcat.views import DcatDatasetCreateView
+
+urlpatterns = [
+    path(
+        "organization/<int:organization_id>/dcat/dataset/create/subclass/<uuid:subclass_uuid>/",
```
Struchid: Food for thought: Do we need to define the `create` inside of the endpoint itself? I guess this would be implied by the HTTP method used for the URL; Ignore if this is something you need for completely different purpose that I do not see here.

vitrina/datasets/form_helpers.py
```diff
@@ -114,3 +128,112 @@ def validate_identifier(identifier: str | None) -> None:
...
+
+    if agent and not cleaned_data.get("endpoint_description_type"):
+        if not (openapi_format := Format.objects.filter(title=OPENAPI_FORMAT).first()):
+            raise ValidationError(error_template.format(JSON_FORMAT))
+        cleaned_data["endpoint_description_type"] = openapi_format
+
+    if agent and not cleaned_data.get("conforms_to"):
+        if not (uapi_concept := Concept.objects.filter(code=UAPI_CONCEPT_CODE).first()):
+            raise ValidationError(error_template.format(JSON_FORMAT))
+        cleaned_data["conforms_to"] = uapi_concept
```
Struchid: I guess this is a case where we could have the outer layer with the condition `if agent` (or a reverse condition with an early return), since the current flow checks for agent 3 times separately;

vitrina/datasets/form_helpers.py
```diff
@@ -114,3 +128,112 @@ def validate_identifier(identifier: str | None) -> None:
...
+    agent: Agent | None,
+    conforms_to: Concept | None,
+    endpoint_url: str | None,
+    endpoint_type: str | None,
+    endpoint_description: str | None,
+    endpoint_description_type: str | None,
+) -> list[tuple[str, str]]:
+    errors = []
+
+    if not agent and not endpoint_url:
```
Struchid: Same here if you decide to follow the additional shrouding layer of `if not agent`

vitrina/datasets/form_helpers.py
```diff
@@ -114,3 +128,112 @@ def validate_identifier(identifier: str | None) -> None:
...
+
+
+def get_contact_form_choices(organization: Organization) -> list[tuple[int, str]]:
+    contact_choices = []
+    content_type_user = ContentType.objects.get_for_model(User)
+    content_type_organization = ContentType.objects.get_for_model(Organization)
+
+    contacts = Contact.objects.filter(organization=organization, deleted__isnull=True).select_related("content_type")
+
+    org_contacts = []
```
Struchid: This might be just moved around, but I would really advise against using abbreviations, they cause a lot of readability issues further down the line

Struchid (approved):
If possible next time, let's at least do two commits (even better - two PRs) for:
a) Moving things around;
b) Small refactorings;

Helps identifying what has changed and what was moved, thus avoiding (me) writing comments about parts that are not implemented by you in this PR :D

vitrina/dcat/urls.py
martynasracct: Oops, these changes are not suppose to be in this PR

vitrina/datasets/form_helpers.py
martynasracct: Yeah, it might be easier to read with single `if agent ... else ...`

Struchid (approved):

PR #2565, opened 2026-04-24
Add DCAT DatasetUpdateView

martynasracct:
**Summary:**
 Adds `DcatDatasetUpdateView`  used for DCAT data update (non-open datasets).

This is part 2 of many PRs related to DCAT data creation. Additional views, forms and form fields will be added later.

DcatDatasetUpdateView is heavily inspired by DatasetUpdateView but used only for DCAT data update: information systems, services and datasets. It is impossible to update open data datasets using this view.

Most likely this will not be merged into devel right away. Most likely we will merge all DCAT related PRs to some dcat-devel branch until we will get green light to release it.

### DIFF
--- a/vitrina/dcat/views.py
+++ b/vitrina/dcat/views.py
@@ -215,8 +215,8 @@ class DcatDatasetUpdateView(
     def subclass(self) -> DCATResourceSubclass:
         return self.get_object().subclass
 
-    def has_permission(self):
-        return has_perm(self.request.user, Action.UPDATE_WIZARD, Dataset, self.organization)
+    def has_permission(self) -> bool:
+        return has_perm(self.request.user, Action.UPDATE_WIZARD, self.get_object())
 
     def dispatch(self, request: WSGIRequest, *args, **kwargs) -> HttpResponseBase:
         obj = self.get_object()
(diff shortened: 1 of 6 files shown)


vitrina/dcat/views.py
```diff
@@ -185,10 +187,179 @@ def form_valid(self, form: BaseResourceForm) -> HttpResponseBase:
...
+
+    @cached_property
+    def organization(self) -> Organization:
+        return get_object_or_404(Organization, pk=self.kwargs.get("organization_id"))
+
+    @cached_property
+    def subclass(self) -> DCATResourceSubclass:
+        return self.get_object().subclass
+
+    def has_permission(self):
```
Struchid: :/ No typehint for return?

Struchid (approved):

vitrina/dcat/views.py
```diff
@@ -185,10 +187,179 @@ def form_valid(self, form: BaseResourceForm) -> HttpResponseBase:
...
+        return has_perm(self.request.user, Action.UPDATE_WIZARD, Dataset, self.organization)
```
eimantas-nejus-cct: Might be totally wrong, but maybe we should be using instance level permission check instead of class+organisation? Like:
```
return has_perm(self.request.user, Action.UPDATE_WIZARD, self.get_object())
```

Otherwise - non-admin users can run into situations where they create a Dataset and immediately are prohibited from updating it? Tests for this scenario are using admin user and admins pass all `has_perm` checks via shortcut.

martynasracct: Good catch! Fixed and added some tests with custom permissions

eimantas-nejus-cct (approved):

PR #2601, opened 2026-05-08
Fix DCAT form field ordering, limit some choices

martynasracct:
**Summary:**
- Add `ConceptOrderedByLabelManager` that orders queryset by `Concept.label` of currently used language. 
- Limit choices of DCAT dataset form `parent` field:
    - `InformationSystem` parent can only be another `InformationSystem` from same organization
    -  `Service` parent can only be `InformationSystem` from same organization
    -  `Dataset` parent can only be `Service` from same organization

### DIFF
--- a/tests/classifiers/test_managers.py
+++ b/tests/classifiers/test_managers.py
@@ -1,0 +1,75 @@
+import pytest
+from django.utils.translation import override as translation_override
+
+from vitrina.classifiers.factories import ConceptFactory
+from vitrina.classifiers.models import Concept
+
+pytestmark = pytest.mark.django_db
+
+
+def _set_label(concept, lang, label, description="desc"):
+    concept.set_current_language(lang)
+    concept.label = label
+    concept.description = description
+    concept.save()
+    return concept
+
+
+class TestConceptOrderedByLabelManager:
+    def test_concepts_ordered_alphabetically_by_translated_label(self):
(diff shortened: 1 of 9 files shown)


tests/classifiers/test_managers.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+from django.utils.translation import override as translation_override
+
+from vitrina.classifiers.factories import ConceptFactory
+from vitrina.classifiers.models import Concept
+
+pytestmark = pytest.mark.django_db
+
+
+def _set_label(concept, lang, label, description="desc"):
```
MindaugasStanionis: typehints are missing

MindaugasStanionis (approved):

tests/classifiers/test_managers.py
```diff
@@ -0,0 +1,75 @@
...
+    concept.description = description
+    concept.save()
+    return concept
+
+
+class TestConceptOrderedByLabelManager:
+    def test_concepts_ordered_alphabetically_by_translated_label(self):
+        c1 = _set_label(ConceptFactory(code="c_zebra"), "lt", "Zebra")
+        c2 = _set_label(ConceptFactory(code="c_apple"), "lt", "Apple")
+        c3 = _set_label(ConceptFactory(code="c_mango"), "lt", "Mango")
```
Struchid: Could we use full variable names?

Struchid (approved):

PR #2627, opened 2026-05-18
Remove duplicated code and rename method

martynasracct:
**Summary:**
- Renames `validate_name_prefix` to `get_name_prefixes`, since function does not validate anything
- Removes duplicated dataset name prefix checks for `OPEN_DATA_PUBLISHER` from `validate_dataset_name`. Same exact code already runs in `get_name_prefixes` function that is called.


Struchid (approved):
Good man 👍

Gustascct (approved):

PR #2641, opened 2026-05-26
2588 statistics fixes

### LINKED ISSUE #2588: Statistics pages keep recalculating on each request
For example, when going to page such https://data.gov.lt/datasets/stats/creator/ we now get 429 status code. This is because we have limited the requests to these endpoints since they were recalculating all stats on each request which slowed the server down a lot.

kastger:
The `/datasets/stats/` pages recompute everything on every request. The main cost was an N+1 query pattern: an aggregation queryset was re-filtered once per time bucket (per month/quarter/year * per series label), so a single page issued hundreds of near-identical queries. On top of that, Haystack iterated the Elasticsearch result set in batches of 10 (the default), turning one logical search into hundreds of ES round-trips.

Loading time before was around 30 seconds (heaviest I got locally, can be tested in test/prod), after these changes it is down to ~2 seconds.


vitrina/requests/views.py
```diff
@@ -435,25 +427,36 @@ def get_context_data(self, **kwargs):
                 period = str(pd.to_datetime(created).to_period(frequency))
                 stats_for_period[period] = stats_for_period.get(period, 0) + 1
         if indicator != "request-count":
+            tz_vilnius = pytz.timezone("Europe/Vilnius")
```
martynasracct: we should use `ZoneInfo` instead of `pytz`. Django 5.0 deprecates `pytz`

vitrina/requests/views.py
```diff
@@ -435,25 +427,36 @@ def get_context_data(self, **kwargs):
...
-                    year_stats[yr] = total
-                else:
-                    total = (
-                        PlanRequest.objects.filter(
-                            request_id__in=request_ids,
-                            plan__deadline__lt=datetime.now(),
-                        )
-                    ).count()
-                    year_stats[yr] = total
+                yr_start = datetime.strptime(str(yr) + "-1-1", "%Y-%m-%d")
```
martynasracct: We should use full names for variables

vitrina/datasets/views.py
```diff
@@ -2513,13 +2531,23 @@ def get_context_data(self, **kwargs):
...
+                id_list_by_display_value[k.get("display_value")] = [dd.pk for dd in cat_datasets]
+                all_dataset_ids.extend(id_list_by_display_value[k.get("display_value")])
+
+            stats_by_dataset_id: dict = {}
+            if indicator not in ("download-request-count", "download-object-count"):
+                all_stats_qs = DatasetStats.objects.filter(dataset_id__in=all_dataset_ids)
+                for st in all_stats_qs:
+                    stats_by_dataset_id.setdefault(st.dataset_id, []).append(st)
+
+            for k in filtered_cats:
```
martynasracct: We should use full names for variables

vitrina/datasets/views.py
```diff
@@ -2752,36 +2810,43 @@ def get_context_data(self, **kwargs):
...
+                            model_stats = ModelDownloadStats.objects.filter(model=m)
+                            if len(model_stats) > 0:
+                                for m_st in model_stats:
+                                    if indicator == "download-request-count":
+                                        if m_st is not None:
+                                            total += m_st.model_requests
+                                    elif indicator == "download-object-count":
+                                        if m_st is not None:
+                                            total += m_st.model_objects
+                    quarter_stats[k] = total
```
martynasracct: I think this could be a query instead of for loops and multiple N+1s. Query all `ModelDownloadStats` once, calculate both totals using `.aggregate(...)` and use one based on `indicator`

vitrina/datasets/views.py
```diff
@@ -2824,15 +2889,18 @@ def get_context_data(self, **kwargs):
                         month = str(year_published) + "-" + str("%02d" % published.month)
                         monthly_stats[month] = monthly_stats.get(month, 0) + 1
         if indicator != "dataset-count":
+            tz_vilnius = pytz.timezone("Europe/Vilnius")
```
martynasracct: Let's use `ZoneInfo` instead of `pytz`

vitrina/datasets/services.py
```diff
@@ -175,6 +176,41 @@ def get_query_for_frequency(frequency, field, label):
     return query
 
 
+def get_period_key(frequency: str, field: str, label: Any) -> tuple:
```
martynasracct, line 178: Can these be changed to `match` cases or to a dictionary?

tests/stats_utils.py
```diff
@@ -0,0 +1,38 @@
...
+from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
+from django.test import RequestFactory
+from django.urls import resolve
+
+
+FROZEN_NOW = "2023-06-15 12:00:00"
+
+
+def get_stats_context(view_class, path, *, params=None, user=None, **view_kwargs):
```
martynasracct: type hints missing

tests/stats_utils.py
```diff
@@ -0,0 +1,38 @@
+import json as _json
```
martynasracct: Is there a reason to import `json as _json`?

martynasracct (approved):

vitrina/datasets/services.py
```diff
@@ -175,6 +176,41 @@ def get_query_for_frequency(frequency, field, label):
...
+    if frequency == "Y":
+        return (label.year,)
+    elif frequency == "Q":
+        return (label.year, label.quarter)
+    elif frequency == "M":
+        return (label.year, label.month)
+    elif frequency == "W":
+        return (label.year, label.month, label.week)
+    else:  # "D"
```
Struchid: Are we expecting only "D" on this line? If not, I'd move this to an `elif` and add an `else` with a warning/error that would be thrown when an unexpected period key is given;

vitrina/datasets/services.py
```diff
@@ -175,6 +176,41 @@ def get_query_for_frequency(frequency, field, label):
...
+def row_period_key(row: Mapping[str, Any], frequency: str, field: str) -> tuple:
+    if frequency == "Y":
+        return (row[f"{field}__year"],)
+    elif frequency == "Q":
+        return (row[f"{field}__year"], row[f"{field}__quarter"])
+    elif frequency == "M":
+        return (row[f"{field}__year"], row[f"{field}__month"])
+    elif frequency == "W":
+        return (row[f"{field}__year"], row[f"{field}__month"], row[f"{field}__week"])
+    else:  # "D"
```
Struchid: Same here

vitrina/datasets/views.py
```diff
@@ -2413,6 +2420,15 @@ def get_context_data(self, **kwargs):
                     id_list.append(d)
             modified[org] = id_list
             filtered_orgs.append(modified)
+
+        all_dataset_ids = [did for single in filtered_orgs for ids in single.values() for did in ids]
```
Struchid: `did` -> `dataset_id`

vitrina/datasets/views.py
```diff
@@ -2628,15 +2657,18 @@ def get_context_data(self, **kwargs):
...
-                start_date = datetime.strptime(str(yr) + "-1-1", "%Y-%m-%d")
-                end_date = datetime.strptime(str(yr) + "-12-31", "%Y-%m-%d")
-                tz = pytz.timezone("Europe/Vilnius")
-                filtered_datasets = datasets.filter(published__range=[tz.localize(start_date), tz.localize(end_date)])
-                dataset_ids = []
-                for fd in filtered_datasets:
-                    dataset_ids.append(fd.pk)
-                if indicator == "download-request-count" or indicator == "download-object-count":
+                yr_start = datetime.strptime(str(yr) + "-1-1", "%Y-%m-%d")
+                yr_end = datetime.strptime(str(yr) + "-12-31", "%Y-%m-%d")
```
Struchid:
```suggestion
                year_start = datetime.strptime(str(year) + "-1-1", "%Y-%m-%d")
                year_end = datetime.strptime(str(year) + "-12-31", "%Y-%m-%d")
```

Struchid (approved):

PR #2651, opened 2026-05-28
Ft/dcat wizard form design polish

eimantas-nejus-cct:
Design overhaul to match the Figma design much closer than before:
  - Used classes for a more streamlined style application instead of inlines and extracted most of it into a dedicated wizard.scss
  - Made fragile positional CSS selectors more robust
  - Greatly reduced number of !important uses
  - Fixed breadcrumb labeling inconsistency issue
  - Consolidated _wizard_messages and _wizard_form_errors partials into a shared template location (used by both orgs and dcat)
  - Fixed N+1 query when rendering breadcrumb ancestors (added prefetch_related for subclass translations)

Examples:

Current:
image

Figma:
image

Implementation:
image


martynasracct (approved):
Looks good!

Struchid (approved):
The design is quite off from what we currently have in this project, was this design overhaul confirmed by VSSA? 😵‍💫 I guess this is only a problem if there are just a few new pages like that and not the whole system

Gustascct (approved):

eimantas-nejus-cct:
> The design is quite off from what we currently have in this project, was this design overhaul confirmed by VSSA? 😵‍💫 I guess this is only a problem if there are just a few new pages like that and not the whole system

Yes, this is intended and the new design is confirmed by the VSSA team. It works, because this is a protected area available only for the organisation admins, where they can quickly create or update the full organisational data and their relations.

PR #2654, opened 2026-05-29
Random UUID inclusion

Gustascct:
Factory was used twice and generated the same word, breaking the unique constraint. Rerunning the tests makes it pass. Would not be that big of a problem, but the pipeline runs for 30 mins 😨 

Pipeline that broke - https://github.com/atviriduomenys/katalogas/actions/runs/26625074385/job/78459744042


vitrina/datasets/factories.py
Struchid, line 1: Wouldn't just using the `fake.uuid4()` be enough? Do you still need the `fake.word()` as well?

Struchid (approved):

MindaugasStanionis (approved):

PR #2661, opened 2026-06-02
Fix datasets not displayed in DCAT wizard path

martynasracct:
**Summary:**
Bug found during VSSA testing: when creating DATASET or SERVICE nodes allows selecting different organization than parent INFORMATION_SYSTEM. In that case, dataset is not displayed in DCAT Wizard tree.

Fix is to also include all datasets with `path` that starts with path of any  INFORMATION SYSTEM in organization


eimantas-nejus-cct (approved):
Nice one!

PR #2662, opened 2026-06-02
Adds DCAT contact form

martynasracct:
**Summary:**

- Adds new field `Contact.kind`. This is a helper field to distinguish between organization, user, unregistered contacts. Previously `ContentType` was used for this purpose. This field is set automatically when creating `Contact` via forms.
- Adds new fields to `Contact` model
- Adds new DCAT forms for creating `Contact`. `Contact` created by DCAT form will always have kind set to "SERVICE". DCAT `Contact` form uses new fields added to the model
- DCAT form is reachable from the same contact page, just requires non-open data coordinator permissions. Contacts created via DCAT view can be seen by open data coordinators in same list


eimantas-nejus-cct:
Could there be a missing migration? 

```
(vitrina-py3.11) ➜  katalogas git:(ft/dcat-contact-form) ✗ poetry run python manage.py makemigrations --check
System check identified some issues:

WARNINGS:
vitrina_users.User: (auth.W004) 'User.email' is named as the 'USERNAME_FIELD', but it is not unique.
        HINT: Ensure that your authentication backend(s) can handle non-unique usernames.
Migrations for 'vitrina_classifiers':
  vitrina/classifiers/migrations/0021_alter_formfieldhelptext_form_name.py
    - Alter field form_name on formfieldhelptext
(vitrina-py3.11) ➜  katalogas git:(ft/dcat-contact-form) ✗ 
```

vitrina/datasets/models.py
```diff
@@ -2264,17 +2264,84 @@ class Meta:
 
 
 class Contact(models.Model):
+    class Kind(models.TextChoices):
+        INDIVIDUAL = "individual", _("Registruotas naudotojas")
+        ORG = "org", _("Organizacija")
+        UNREGISTERED = "unregistered", _("Neregistruotas naudotojas")
+        SERVICE = "service", _("Paslauga")
```
Struchid: I'd prefer this would be defined in the __init__ file rather than the model itself

Struchid (approved):

PR #2664, opened 2026-06-03
Revert "DCAT - Design fixes and improvements"

eimantas-nejus-cct:
Webpack is not being rebuilt in Review App redeployment, where testing is currently ongoing, so we need to revert changes so VSSA team is unblocked and can continue testing the logic of the forms. This change will be built as a separate review app


martynasracct (approved):

eimantas-nejus-cct:
False alarm, needed to hard refresh the page so it grabs the latest css. Closing the PR.

PR #2668, opened 2026-06-04
[DCAT-20] Add hasQualityAnnotation

Struchid:
Adding a new field to Dataset forms - `hasQualityAnnotation`;

DCAT reference:
- https://www.w3.org/TR/vocab-dqv/#dqv:hasQualityAnnotation
- https://ivpk.github.io/DCAT-AP-LT/#%23%23PasirenkamosDuomen%C5%B3RinkinysSavybes

### Changes:

- `QualityAnnotationAdminForm` added to validate that one of dataset/distribution is selected (not both) according to the comment:
> Pastaba: Vienas kokybės anotacijos egzempliorius gali vertinti tik duomenų rinkinius arba tik pateiktis, ne abu. Privalomas yra bent vienas iš jų."

- Two new models `QualityAnnotation`, `QualityAnnotationBody`; The model with `...Body` is needed to assign multiple names to `QualityAnnotation` due to the requirements listed in the xlsx file;
- Text field is chosen for the `QualityAnnotationBody` since the value can be either an URL or a comment:
> Laisvu tekstu pateikiamas vertinimo komentaras arba nuoroda į  vertinimo failą."


MindaugasStanionis (approved):

vitrina/datasets/models.py
```diff
@@ -2438,3 +2445,62 @@ class DatasetExcludedGroups(models.Model):
     class Meta:
         unique_together = ("dataset", "group")
         db_table = "dataset_excluded_groups"
+
+
+class QualityAnnotationBody(models.Model):
```
martynasracct: Let's use `UUIDBaseModel` as parent

martynasracct (approved):

vitrina/locale/en/LC_MESSAGES/django.po
eimantas-nejus-cct, line 1: Shouldn't the LT translations be generated as well so they are in sync? I know they're empty, maybe it's a convention in the project?

vitrina/datasets/models.py
```diff
@@ -2438,3 +2445,62 @@ class DatasetExcludedGroups(models.Model):
...
+    value = models.TextField(
+        verbose_name="Reikšmė",
```
eimantas-nejus-cct: Missing translation wrapper
```suggestion
        verbose_name=_("Reikšmė"),
```

eimantas-nejus-cct (approved):

vitrina/datasets/models.py
Struchid: ... Regenerating migrations 😢 Done!

vitrina/datasets/models.py
Struchid: Nice eye, fixed!

vitrina/locale/en/LC_MESSAGES/django.po
Struchid: Not sure, but sure, added.

vitrina/datasets/models.py
```diff
@@ -2438,3 +2445,62 @@ class DatasetExcludedGroups(models.Model):
...
+
+class QualityAnnotation(UUIDBaseModel):
+    codename = models.CharField(
+        max_length=255,
+        unique=True,
+        verbose_name=_("Kodinis pavadinimas"),
+        help_text=_(
+            "Laisvu tekstu įvedamas kodinis pavadinimas mažosiomis lotyniškomis raidėmis. "
+            "Pradinio duomenų surinkimo palengvinimui. Šiame stulpelyje negali būti pasikartojančių reikšmių. "
+            "Formatas: kod_pav"
```
Gustascct, line 2474: Just making sure, the format is actually kod_pav? To me it seems that it does not refer to anything

Gustascct (approved):

vitrina/datasets/models.py
Struchid: It does not have to refer to anything and is more of a helper for the user trying to fill out the field value, helping him understand the format (if the description is not enough).

Either way these texts are taken directly from the xlsx file I was given for the task, so just copied the description of the field from there:

image

PR #2674, opened 2026-06-05
[DCAT-62] Extend phone validator to organization form

Struchid:
Extend the phone validator to organization forms:

image

### DIFF
--- a/tests/orgs/test_forms.py
+++ b/tests/orgs/test_forms.py
@@ -45,6 +45,22 @@ class TestOrganizationCreateForm:
         generated_name = form.cleaned_data["name"]
         assert generated_name == "datasets/gov/test-org/"
 
+    def test_phone_invalid_fails_validation(self, app: DjangoTestApp):
+        user = UserFactory()
+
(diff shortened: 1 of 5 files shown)


tests/orgs/test_forms.py
```diff
@@ -34,6 +34,24 @@ def test_organization_update_form_disables_name_for_existing_instance(self, app:
         assert "disabled" in form.fields["name"].widget.attrs
 
 
+class TestOrganizationBaseForm:
+    def test_phone_invalid_fails_validation(self, app: DjangoTestApp):
+        user = UserFactory()
```
Gustascct, line 49: Is it better to have a separate class for these tests? Maybe it would be better to put the tests in the OrganizationCreateForm class?

Gustascct (approved):
Gytis išties šaunuolis

tests/orgs/test_forms.py
Struchid: Yup, might have been to eager to trust some friend who suggested otherwise :-)

martynasracct (approved):

PR #2676, opened 2026-06-05
DCAT wizard: Improve intuitiveness of the wizard navigation

eimantas-nejus-cct:
Improve organisation wizard intuitiveness — make it clear which element types can be created at which hierarchy level.

### Changes:

  - **"Kūrimo tvarka" legend** — added a collapsible matrix table at the top of the wizard sidebar showing which child types each parent type can have;
  - **Stable type picker** — the create screen now always lists all five element types in a fixed order, but only allowed elements are selectable;
  - **Consistent terminology** — unified "mazgas" / "objektas" / "įrašas" wording to "elementas" across wizard helpers, breadcrumbs and tooltips;
  - **Single source of truth for type config**;
  - **Pateiktis form cleanup** — removed the leftover inline "Redaguoti"/"Sukurti" button;

Example:
image

### DIFF
--- a/vitrina/locale/en/LC_MESSAGES/django.po
+++ b/vitrina/locale/en/LC_MESSAGES/django.po
@@ -2781,17 +2781,17 @@ msgstr "Distributions"
 msgid "E. paslaugos"
 msgstr "E-services"
 
-msgid "Sukurti naują informacinę sistemą po šiuo mazgu."
-msgstr "Create new information system under this node."
+msgid "Sukurti naują informacinę sistemą po šiuo elementu."
+msgstr "Create a new information system under this element."
 
-msgid "Sukurti naują e. paslaugą po šiuo mazgu."
-msgstr "Create a new e-service under this node."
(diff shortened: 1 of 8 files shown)


vitrina/locale/en/LC_MESSAGES/django.po
```diff
@@ -7395,3 +7404,9 @@ msgid ""
 "Prisijunkite su šiuo elektroniniu paštu ir bandykite viisp autentifikaciją "
 "iš naujo"
 msgstr "Log in with this e-mail address and try viisp authentication again"
+
+#~ msgid "Pasirinkite objektą"
+#~ msgstr "Select an object"
+
+#~ msgid "Šiam mazgui nėra leidžiamų vaikinių elementų."
+#~ msgstr "This node has no allowed child types."
```
Struchid, line 7545: These should be removed

Struchid (approved):

Gustascct (approved):

martynasracct (approved):

PR #2683, opened 2026-06-08
[DCAT-47] Make `availability` a non-required field

Struchid:
`availability` is only required when given (meaning optional);

According to the document:
image


martynasracct (approved):
Does this automatically add empty choice in form?

Struchid:
> Does this automatically add empty choice in form?

The widget allows to remove current selections:
image

The field can be kept empty:
image

No additional choice for `empty` is needed

Gustascct (approved):

PR #2684, opened 2026-06-08
Change distribution status choices

martynasracct:
**Summary:**
Previously `DatasetDistribution.status` choices were taken from `http://publications.europa.eu/resource/authority/distribution-status`. According to DCAT, choices should be taken from `http://purl.org/adms/status/`.

So, in DCAT forms, choices for distributions will have choices from `http://purl.org/adms/status/`. Already existing records, that have old status choice already set will still display it (and keep it after save if not changed), but it will only be possible to change it to one of the new values.

Additionally, updated Dataset codename labels and help_texts


Struchid (approved):

gts446 (approved):
Shouldn't there also be a migration to populate new choices? Or they stay the same just URI is different? In that case, shouldn't there be a migration to update URI? :)

martynasracct:
Both `ConceptSchema` and `Concept` already exists in database (already created via migration) because they are used in another field. So I only needed to change the queryset

Struchid (approved):

PR #2685, opened 2026-06-08
[DCAT-53] Field `access_rights` adjustments

Struchid:
Adjustments according to the DCAT excel and comments on the fixes excel.

Changes:
- Field is not required;
- Default is removed;

image

image


Gustascct (approved):

vitrina/datasets/models.py
```diff
@@ -358,8 +358,8 @@ class Dataset(Resource):
 
     access_rights = models.CharField(
         _("Prieigos teisės"),
+        null=True,
         blank=True,
-        default=PUBLIC,
```
martynasracct, line 362: Can we still keep default value for non DCAT forms?

martynasracct (approved):

vitrina/datasets/models.py
Struchid: I'd say it is not really needed since forms DCAT or not should conform to the excel and not have a default set if there is no value given, but let's discuss this in PMs

PR #2688, opened 2026-06-10
Ft/dcat add missing fields

martynasracct:
**Summary:**
- Renames SERVICE field "Kodinis pavadinimas" to "Identifikatorius"
- Adds new field "relates_to_data_service" to DCAT INFORMATION_SYSTEM form
- Renames "Tinklapis" to "Tinklalapis"


vitrina/locale/en/LC_MESSAGES/django.po
```diff
@@ -2419,12 +2419,26 @@ msgstr ""
 msgid "Aprašas"
 msgstr "Description"
 
-msgid "Tinklapis"
-msgstr "Endpoint"
```
eimantas-nejus-cct, line 2423: Optional. "Tinklalapis" is translated as "Website" already, so in English translation the value will change from Endpoint -> Website. DCAT field is named dcat:endpointURL, so maybe worth checking it out? Although changing it to "Endpoint" will effect places where it is currently using this translation.. Anyways, minor issue, I guess, but wanted to point it out.

eimantas-nejus-cct (approved):

vitrina/locale/en/LC_MESSAGES/django.po
martynasracct: Yes, I saw that, but I will keep it as is for now.

PR #2710, opened 2026-07-02
2585 Fix VIISP registration 500 error and company code assignment

### LINKED ISSUE #2585: Peržiūrėti registracijos per VIISP procesą
Reikia peržiūrėti registracijos per VIISP procesą, nes aš asmeniškai bandžiau su savo paskyra, tai man metė gal 2-3 klaidas visame procese, bet galiausiai pavyko susikurti userį, o yra paklausimų, kurie taip pat negali prisiregistruoti per VIISP, meta 500 klaidą, tai reiktų peržiūrėti procesą ir ištaisyti klaidas.

kastger:


karina-klinkeviciute:
● Code review(high · 9 findings)                                                                                                                                                                                             
  ⎿  ●vitrina/viisp/views.py:[correctness— A failed bcrypt personal-code check in the elif user_social_account branch falls through to a fresh social login instead of rejecting, because every return sits inside the       
      117                    ]           if bcrypt.checkpw(...) block with no else.                                                                                                                                          
       Existing user with a SocialAccount logs in without a token and the VIISP personal_code does not match the stored bcrypt hash → checkpw returns False → nothing returns → execution reaches the unconditional          
       provider.sociallogin_from_response / complete_social_login tail (~line 144) and logs into the account owning the VIISP-supplied email despite the failed identity check (fail-open).                                  
     ●vitrina/viisp/xml_utils.py[correctness— The new try/except only wraps resp.raise_for_status(); the post(VIISP_PROXY_AUTH, ...) call itself is outside the guard, so ConnectionError/Timeout still propagate as         
      :110                      ]           unhandled 500s — the exact outage the PR claims to fix.                                                                                                                          
       VIISP proxy is unreachable or times out → requests.post() raises ConnectionError/Timeout (both RequestException subclasses) before raise_for_status is reached → exception escapes get_response_with_user_data →      
       500 instead of the api_error page. The sibling get_response_with_ticket_id has the identical mis-scoped structure, and _parse_user_data can also raise on a malformed SOAP body outside any guard.                    
     ●vitrina/viisp/views.py:[correctness— ViispTokenKey.objects.first().key_content is still unguarded in _confirm_viisp_email (line 191) and VIISPAccountMergeView.get (line 226), so the missing-key 500 the PR           
      191                    ]           claims to fix survives on these paths.                                                                                                                                              
       ViispKey seeded but ViispTokenKey table empty; an existing user with matching email but no SocialAccount completes VIISP login without a token → post() falls to the else branch → _confirm_viisp_email →             
       .first() returns None → AttributeError → HTTP 500. Same crash on the account-merge link via VIISPAccountMergeView.get.                                                                                                
     ●vitrina/viisp/views.py:11[correctness]— user_social_account.extra_data.get("personal_code").encode("utf-8") dereferences None for SocialAccounts created before personal_code was adde added to extra_data (pre-commit      
      9                                     07a81fa0 rows).                                                                                                                                                                  
       A user who registered via VIISP before personal_code was stored logs in without a token → user_social_account found (the lookup is not even provider-scoped) → .get("personal_code") returns None →                   
       AttributeError → 500. The new guard at line 92 only checks the incoming VIISP data, not the persisted legacy extra_data.                                                                                              
     ●vitrina/viisp/views.py:8[correctness]— fernet.decrypt(token) is unguarded on a csrf_exempt view whose token comes from the URL path, so any malformed or tampered token raises cryptography.fernet.InvalidToken →      
      3                                    unhandled 500.                                                                                                                                                                    
       Anyone requests viisp/complete-login/<garbage> with a valid ticket → decrypt raises InvalidToken (or binascii.Error) → 500, while the adjacent lines were just hardened against missing keys; an                      
       attacker-triggerable 500 remains.                                                                                                                                                                                     
     ●vitrina/viisp/views.py:84 [correctness] — The token-flow email check `email != user_data.get("email")` is case-sensitive while neither side is normalized, so a casing difference wrongly redirects to                 
                                              change-email.                                                                                                                                                                  
       User.email stored with mixed case (CharField, no lowercasing in registration) while VIISP returns the same address in different casing (parse path does not lowercase; provider lowercasing happens only later insociallogin_from_response) → same account is treated as an email mismatch and bounced to change-email during the token flow.                                                                                          
     ●vitrina/viisp/xml_utils.py[observabilit— The RequestException is swallowed with no logging and the views render api_error.html with an empty {} context and HTTP 200, so VIISP outages are invisible to logs,          
      :113                      y]           monitoring, and the template's superuser diagnostic block.                                                                                                                      
       VIISP returns an HTTP error → get_response_with_user_data silently returns None (no logger call, though getLogger is used in 17 vitrina files) → api_error.html rendered with {} so                                   
       error_response/error_response_text are blank → status 200 means alerting never fires and the outage is undebuggable.                                                                                                  
     ●vitrina/viisp/views.py:[reuse]— The diff adds five more verbatim `return render(request, "allauth/socialaccount/api_error.html", ...)` calls, bringing the file to seven copies (lines 47, 53, 58, 71, 77, 81, 94)     
      47                            with no shared helper; the key-fetch-then-guard blocks are likewise copy-pasted.                                                                                                         
       Changing the template path, status code, or adding diagnostic context requires editing seven call sites; a small helper (e.g. _viisp_api_error(request, error_data=None) plus a                                       
       get_viisp_key()/get_viisp_fernet() accessor) would collapse the duplication and would also have covered the still-unguarded sites at lines 191/226.                                                                   
     ●vitrina/viisp/views.py:9[simplification]— The proxy_type → viisp_company_code assignment block is duplicated verbatim at lines 99-102 and 152-155; this exact duplication is what produced the user-vs-login.user      
      9                                       bug the PR fixes.                                                                                                                                                              
       The next change to the proxy-type rule can again be applied to only one copy and silently diverge; extracting a _set_company_code(target_user, user_data) helper removes the risk.

karina-klinkeviciute (approved):

PR #2720, opened 2026-07-09
2719 spinta request

### LINKED ISSUE #2719: Klaida teikiant užklausą kataloge Duomenys skiltyje
Naudotojo paklausimas:

Siunčiant užklausą: https://data.gov.lt/datasets/2088/versions/1605/data/JuridinisAsmuo/?select(_id,kodas,pavadinimas,isregistruotas,veikla,iregistruotas)&format(csv)

Gaunama klaida:

UnexpectedToken
Unexpected token Token('AMPERSAND', '&') at line 1, column 67. Expected one of: * SIGN * LPAR * NAME * STRING * FLOAT * SORT * NULL * BOOL * ALL * COUNT * SELECT * INT * LSQB * LIMIT Previous tokens: [Token('COMP', '=')]

Galimai problema yra nuorodos encode'inime, kurios paskui nesupranta saugykla iį kurią kreipiamasi ir meta klaida, pvz.:
https://get.data.gov.lt/datasets/gov/lsd/cl/ja_asmenys/JuridinisAsmuo/:format/csv?select(_id,kodas,pavadinimas,isregistruotas,veikla,iregistruotas) veikia, bet puslapis padaro tokia užklausa: https://get.data.gov.lt/datasets/gov/lsd/cl/ja_asmenys/JuridinisAsmuo/:format/csv?select%28_id%2Ckodas%2Cpavadinimas%2Cisregistruotas%2Cveikla%2Ciregistruotas%29=

kastger:
Filtering data from Saugykla:
https://data.gov.lt/datasets/2088/versions/1605/data/JuridinisAsmuo/?select(_id,kodas,pavadinimas,isregistruotas,veikla,iregistruotas)

image

And trying to download it after filtering:
image
image

Or the other error in the original task related to this branch

### DIFF
--- a/vitrina/structure/views.py
+++ b/vitrina/structure/views.py
@@ -86,6 +86,7 @@ from vitrina.structure.services import (
     transform_coordinates,
     get_data_from_spinta_async,
     get_allowed_visibilities,
+    build_rql_query,
 )
 from vitrina.datasets.structure import read as read_structure
 from vitrina.structure.utils import TYPE_CHECKER_MAP
@@ -846,19 +847,8 @@ class PropertyStructureView(
 
 async def get_model_data(request, *args, **kwargs):
     model = kwargs.get("model", "").replace("-", "/")
-    query = ["limit(100)"]
-    for key, val in request.GET.items():
-        if key.startswith("select("):
-            select = key
-            query.append(select)
-        else:
-            if val == "":
-                query.append(key)
-            else:
-                tag = f"{key}={val}"
-                query.append(tag)
-
-    query = "&".join(query)
+    query = build_rql_query(request.GET)
+    query = f"limit(100)&{query}" if query else "limit(100)"
     data = await get_data_from_spinta_async(model, query=query)
     return JsonResponse(data)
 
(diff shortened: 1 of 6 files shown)


karina-klinkeviciute (approved):

vitrina/structure/views.py
karina-klinkeviciute, line 1542: It still hand-rolls the query string, so a filter value containing & produces a broken model-data link. This is the same defect class the PR fixed for the download path, 
  and build_rql_query(self.request.GET) would fix it too.

karina-klinkeviciute: It's the same loop build_rql_query now replaces — minus the encoding. self.request.GET.items() gives Django-decoded values, and this loop re-joins them with & without re-encoding. So any &, space, or non-ASCII character
  that lives inside a filter value gets emitted raw.                                                                                                                                                                         
                                                                                                                                                                                                                             
  Concrete case — the user filters prop="A&B" (URL ?prop=%22A%26B%22). Django decodes val to "A&B", and the loop produces:                                                                                                   
                                                                                                                                                                                                                             
  prop="A&B"                                                                                                                                                                                                                 
                                                                                                                                                                                                                             
  The displayed example URL becomes …/Model?prop="A&B". That raw & is a query-param delimiter to Spinta, so if the user actually copies and runs the example, the server parses it as two params (prop="A and B") — the      
  example request is wrong for the very endpoint the tab exists to demonstrate. build_rql_query would encode it to prop="A%26B", which round-trips correctly.

kastger: Fixed

kastger: Fixed, added a test.

PR #2727, opened 2026-07-15
2267 Fix distribution CSV preview and add fullscreen toggle

### LINKED ISSUE #2267: Duomenų šaltinio peržiūra negalima
Pvz rinkinys https://data.gov.lt/datasets/1194/

Paspaudus "Peržiūrėti" jokių duomenų neatvaizduoja, nors duomenys pateikti CSV formatu.

Taip pat, pvz šitam rinkiniui https://data.gov.lt/datasets/1002/ paspaudus "Peržiūrėti" kažkodėl atidaro langą su komentaro ištrynimo opcija, kas yra visiškai nesusiję.

Image

Image

Jeigu paspaudžiu ištrinti, tai ištrina žemiau pateiktus komentarus kol nebėra ko trinti.

kastger:
image
image

### DIFF
--- a/vitrina/datasets/templates/vitrina/datasets/detail.html
+++ b/vitrina/datasets/templates/vitrina/datasets/detail.html
@@ -310,6 +310,43 @@
 {% block scripts %}
     {% load static %}
     {% get_hit_count_js_variables for dataset as hitcount %}
+    <style nonce="{{ request.csp_nonce }}">
+        #preview-modal .modal-card.is-fullscreen {
+            position: fixed;
+            top: 0;
+            left: 0;
+            width: 100vw;
+            width: 100dvw;
+            height: 100vh;
+            height: 100dvh;
+            max-width: 100vw;
+            max-width: 100dvw;
+            max-height: 100vh;
+            max-height: 100dvh;
+            margin: 0;
+            border-radius: 0;
+        }
+        .cms-toolbar-expanded #preview-modal .modal-card.is-fullscreen {
+            top: 46px;
+            height: calc(100vh - 46px);
+            height: calc(100dvh - 46px);
+        }
+        #preview-content {
+            padding-top: 0;
+        }
+        #info-table-body thead th {
+            position: sticky;
+            top: 0;
+            z-index: 1;
+            background-color: #fff;
+            border-bottom: 2px solid #dbdbdb;
+        }
+        @media screen and (max-width: 480px) {
+            .preview-fullscreen-toggle .fullscreen-label {
+                display: none;
+            }
+        }
+    </style>
     <script type="text/javascript" nonce="{{ request.csp_nonce }}">
         $(document).ready(function () {
             $.postCSRF("{{ hitcount.ajax_url }}", { hitcountPK : "{{ hitcount.pk }}" });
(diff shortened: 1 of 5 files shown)


vitrina/datasets/views.py
```diff
@@ -724,38 +725,49 @@ def get(self, request):
...
                     data = [["Only one sheet is allowed in file"]]
                 else:
-                    read_data = read_data[next(iter(read_data))]
-                    headers = read_data.columns.values
-                    first_5_values = read_data.values.tolist()
-                    data.append(list(headers))
-                    data.extend(first_5_values[:5])
+                    read_data = excel_file.parse(excel_file.sheet_names[0], nrows=row_limit)
+                    data.append(list(read_data.columns.values))
+                    data.extend(read_data.values.tolist())
```
karina-klinkeviciute: data.extend(read_data.values.tolist()) emits Python nan for missing cells, and JsonResponse serializes it as the bare token NaN, which is invalid JSON. jQuery $.getJSON (JSON.parse, strict since jQuery 3) throws on it, 
  so the loadPreview success callback never runs and the table renders blank. I confirmed the output: {"data": [["a","b","c"],[1.0,NaN,3.0]]}. Empty cells are extremely common in real CSVs, so this silently blanks many   
  previews. Pre-existing, but the diff re-touches these lines and now reads more rows, widening exposure. Fix: read_data = read_data.where(pd.notnull(read_data), None) (or .fillna("")) before .tolist().

vitrina/datasets/views.py
```diff
@@ -724,38 +725,49 @@ def get(self, request):
...
                     except ValueError:
-                        pass
+                        continue
                 if read_data is not None:
-                    headers = read_data.columns.values
-                    first_5_values = read_data.values.tolist()
-                    data.append(list(headers))
-                    data.extend(first_5_values[:5])
+                    data.append(list(read_data.columns.values))
+                    data.extend(read_data.values.tolist())
```
karina-klinkeviciute, line 772: data.extend(read_data.values.tolist()) emits Python nan for missing cells, and JsonResponse serializes it as the bare token NaN, which is invalid JSON. jQuery $.getJSON (JSON.parse, strict since jQuery 3) throws on it, 
  so the loadPreview success callback never runs and the table renders blank. I confirmed the output: {"data": [["a","b","c"],[1.0,NaN,3.0]]}. Empty cells are extremely common in real CSVs, so this silently blanks many   
  previews. Pre-existing, but the diff re-touches these lines and now reads more rows, widening exposure. Fix: read_data = read_data.where(pd.notnull(read_data), None) (or .fillna("")) before .tolist().

vitrina/datasets/templates/vitrina/datasets/detail.html
```diff
@@ -321,22 +358,49 @@ <h2 class="custom-title is-size-4-mobile no-margin-bottom">{% translate "Priskir
...
+            }
+
+            $('.preview-button').click(function () {
+                currentPreviewId = $(this).attr('id');
+                $('#preview-modal').addClass('is-active');
+                $('html').addClass('is-clipped');
+                loadPreview(currentPreviewId, 5);
             });
 
             $('.new-preview-button').click(function () {
```
karina-klinkeviciute, line 406: currentPreviewId is only cleared in closePreview() and only set by .preview-button. If a user opens a CSV preview, then (without the modal being closed via closePreview) triggers the structure preview flow,             
  currentPreviewId is still the old distribution id. Clicking the fullscreen toggle then calls loadPreview(currentPreviewId, …), replacing the structure table with stale CSV rows. Set currentPreviewId = null; at the top  
  of the .new-preview-button handler.

karina-klinkeviciute, line 406: 3. .new-preview-button doesn't clear the table before inserting rows — detail.html:406-410                                                                                                                                 
  Unlike the old .preview-button handler (which called $("#info-table-body").empty()), this path inserts rows without first emptying #info-table-body. It relies entirely on closePreview() having run. If the modal is      
  reused without a full close (see above), rows append onto existing content. Add an explicit table_obj.innerHTML = '' before the loop.

karina-klinkeviciute (approved):

vitrina/datasets/views.py
kastger: Fixed with `.fillna("")`- `.where(..., None)` forces back to `NaN` on numeric columns.

vitrina/datasets/views.py
kastger: Fixed

vitrina/datasets/templates/vitrina/datasets/detail.html
kastger: Fixed

kastger: Fixed

PR #2731, opened 2026-07-21
[No ticket] Implement custom regex validation

Struchid:
Implement translations for custom regex field validation;

The current approach is not enough - it only shows that there is a problem with an entered value and the regex; We need a custom field that helps the administrator give the user information on what is expected in the field (sort of an explanation for the regex itself);

### DIFF
--- a/vitrina/identifiers/admin.py
+++ b/vitrina/identifiers/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from parler.admin import TranslatableAdmin
 
 from .models import Identifier, Agency
 from vitrina.admin import RevisionCommentVersionAdmin
@@ -50,7 +51,7 @@ class IdentifierAdmin(RevisionCommentVersionAdmin):
 
 
 @admin.register(Agency)
-class AgencyAdmin(RevisionCommentVersionAdmin):
+class AgencyAdmin(TranslatableAdmin, RevisionCommentVersionAdmin):
     list_display = ("name", "uri")
     search_fields = ("name", "uri")
     inlines = [IdentifierInline]
--- a/vitrina/identifiers/factories.py
+++ b/vitrina/identifiers/factories.py
@@ -11,6 +11,7 @@ class AgencyFactory(DjangoModelFactory):
     uri = "http://registrai.lt"
     identifier_validation_type = "REGEXP"
     identifier_validation_options = r"^\d{4}$"
+    identifier_validation_error_message = "Žymėjimas turi būti sudarytas iš keturių skaitmenų."
 
     class Meta:
         model = Agency
--- a/vitrina/identifiers/migrations/0006_agency_identifier_validation_error_message.py
+++ b/vitrina/identifiers/migrations/0006_agency_identifier_validation_error_message.py
@@ -1,0 +1,127 @@
+from django.conf import settings
+from django.db import migrations, models
+
+import parler.fields
+import parler.models
+
+
+FOUR_DIGIT_PATTERN = r"^\d{4}$"
+FOUR_DIGIT_MESSAGE = "Žymėjimas turi būti sudarytas iš keturių skaitmenų."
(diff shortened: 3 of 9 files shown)


vitrina/identifiers/admin.py
```diff
@@ -50,7 +51,22 @@ def get_queryset(self, request):
 
 
 @admin.register(Agency)
-class AgencyAdmin(RevisionCommentVersionAdmin):
+class AgencyAdmin(TranslatableAdmin, RevisionCommentVersionAdmin):
     list_display = ("name", "uri")
     search_fields = ("name", "uri")
     inlines = [IdentifierInline]
+    fieldsets = (
```
martynasracct: Why is this needed?

vitrina/identifiers/factories.py
```diff
@@ -15,6 +16,19 @@ class AgencyFactory(DjangoModelFactory):
     class Meta:
         model = Agency
 
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
```
martynasracct: Why is this needed?

vitrina/identifiers/migrations/0006_agency_identifier_validation_error_message.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("vitrina_identifiers", "0005_add_pasis_agency"),
+    ]
+
+    operations = [
+        MakeAgencyTranslatableInState(),
```
martynasracct, line 87: I don't understand what this does and why it's needed. Is it something that Django generates automatically?

vitrina/identifiers/factories.py
Struchid: My bad, this is a leftover when I tried to seed the english translations together with the lithuanian ones, but we do not care about this in tests, I will remove it

vitrina/identifiers/admin.py
Struchid: Not needed, without it, the fields are in order as I'd like as well (following the model definitions), removing

vitrina/identifiers/migrations/0006_agency_identifier_validation_error_message.py
Struchid: Yeah, I would not think about this myself; But I got an error:
`TypeError: Translatable model does not appear to inherit from TranslatableModel`

So after discussing with my friend, he explained we need this because Django's migration system doesn't use the live `models.py` when applying migrations and uses a kind of a fake "historical" Agency from prior migration operations. That fake Agency has models.Model as base and not TranslatableModelMixin, Model (as with this update);

When the new model `AgencyTranslation` gets created, its FK (parler.fields.TranslationsForeignKey) inspects the fake Agency and refuses to attach unless it inherits `TranslatableModelMixin`

So we have to in a sense fake Agency in migration state to tell it that it is translatable. Without this change - parler raises `TypeError`.

And also a confirmation from our sponsors :P

```
Exactly right. If Agency had been created with bases=(TranslatableModelMixin, models.Model) in 0001_initial.py (like StatRoute, DatasetGroup, DCATResourceSubclass all were), the historical Agency would
  already carry the mixin, and today's migration could just add AgencyTranslation — no workaround needed.

  Every other translatable model in this codebase was born translatable from migration 0001, which is why you don't see any of them using state-manipulation tricks. Ours is retrofitting an existing plain
  model, and Django 4.2 has no first-class AlterModelBases operation, so the little custom MakeAgencyTranslatableInState is filling that gap.

  So: this cost is a one-time price for converting an already-shipped model. Once this migration is applied, all future migrations for Agency (add another translated field, etc.) work normally without any
  state trickery.
```

martynasracct: I see. Makes sense!

martynasracct (approved):

PR #2738, opened 2026-07-21
Do not show form save buttons in DCAT wizard new DCAT element selecti…

martynasracct:
**Summary:**
Buttons "Ištrinti", "Atšaukti pakeitimus" and "Išsaugoti" were showed even when selecting what new element to add:
image

After the fix, buttons are no longer showed in this page:
Screenshot 2026-07-21 at 16 10 15

### DIFF
--- a/vitrina/orgs/templates/vitrina/orgs/wizard.html
+++ b/vitrina/orgs/templates/vitrina/orgs/wizard.html
@@ -261,7 +261,7 @@
                    onclick="event.preventDefault(); wizardNavigateTo(this.href);">
                     {% translate "Uždaryti" %}
                 </a>
-                <div class="buttons">
+                <div class="buttons" x-show="mode !== 'create'">
                     <div id="wizard-footer-extra"></div>
                     <button type="button" class="button is-light"
                             id="wizard-cancel-btn"


vitrina/orgs/templates/vitrina/orgs/wizard.html
```diff
@@ -261,7 +261,7 @@ <h2 class="wizard-picker-title">{% translate "Pasirinkite ką norite sukurti" %}
                    onclick="event.preventDefault(); wizardNavigateTo(this.href);">
                     {% translate "Uždaryti" %}
                 </a>
-                <div class="buttons">
+                <div class="buttons" x-show="mode !== 'create'">
```
Struchid, line 264: Not sure what this x-show does, but you do not need either update/delete here as well, right?

Struchid (approved):

vitrina/orgs/templates/vitrina/orgs/wizard.html
martynasracct: No. There are two types used currently: `create` and `detail`. These types are used to distinguish if displayed page is a form for saving object or is it a page for selecting what type of resource to create. `x-show` controls if element with delete/close/save buttons should be showed or not.

Similar logic is already used in this same file for other elements

Struchid: Oh ok, thanks for the explanation!

PR #2756, opened 2026-07-29
dvms#262: Prevent browser caching for authenticated users

sirex:
**Prevent browser caching for authenticated users**

Adds `NoCacheMiddleware` that sets `Cache-Control: no-cache, no-store, must-revalidate` and `Pragma: no-cache` headers on all responses served to authenticated users, preventing sensitive data from being cached in the browser. The middleware is placed after `AuthenticationMiddleware` so `request.user` is available.

Includes tests (`tests/security/test_cache_headers.py`) that verify:
- All common paths receive no-cache headers for authenticated users
- Anonymous users do not receive these headers (existing caching behavior preserved)

Fixes atviriduomenys/dvms#262

### DIFF
--- a/tests/security/test_cache_headers.py
+++ b/tests/security/test_cache_headers.py
@@ -1,0 +1,48 @@
+from unittest.mock import Mock
+
+import pytest
+from django.http import HttpResponse
+from django.test import RequestFactory
+from django.contrib.auth.models import AnonymousUser
+
+from vitrina.middleware import NoCacheMiddleware
+
+
+ALL_PATHS = [
+    "/datasets/",
+    "/projects/",
+    "/requests/",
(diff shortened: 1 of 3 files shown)


kastger:
Also, it would be better to split changes related to caching and nix into two separate branches as they are completely unrelated to each other. One could stall the other.

sirex:
> Also, it would be better to split changes related to caching and nix into two separate branches as they are completely unrelated to each other. One could stall the other.

As requested, I split the Nix part to a separate pull request:

- https://github.com/atviriduomenys/katalogas/pull/2757

Now this PR only contains changes related to caching.

JPriev (approved):

PR #2757, opened 2026-07-30
Add Nix dev env and update project docs/set up

### LINKED ISSUE #2745: Poetry version mismatch causes `poetry install` to exit with code 1
Running `docker compose up -d` fails during the Docker image build at the `poetry install --no-interaction --only main` step (Step 12/19 in `docker/Dockerfile`).

https://github.com/atviriduomenys/katalogas/blob/efeae5ae7c399aa7d2806e4d60f1eb58484a32e1/docker/Dockerfile#L43-L46

**Error**:

> The command '/bin/sh -c poetry install --no-interaction --only main' returned a non-zero code: 1

**Root cause**:

`poetry.lock` was generated by **Poetry 2.2.1**, but the Dockerfile installs the latest Poetry via `pip install poetry`, which currently resolves to **Poetry 2.4.1**. The version mismatch causes Poetry to re-resolve dependencies and rewrite the lock file during install, which then fails.

**Relevant files:**

- `docker/Dockerfile` — line 44: `RUN pip install gunicorn==20.1.0 poetry --no-cache-dir`
- `poetry.lock` — line 1: header states `Poetry 2.2.1`

**Fix:**

Pin the Poetry version in the Dockerfile to match the lock file:

```dockerfile
RUN pip install gunicorn==20.1.0 poetry==2.4.1 --no-cache-dir
```

Or regenerate the lock file with the latest Poetry version (`poetry lock`) and commit the updated `poetry.lock`.

sirex:
- Add `flake.nix`/`flake.lock` for reproducible Nix dev shell for system packages
- Add `AGENTS.md` with AI context, spec mappings, and dev guide
- Rewrite `README.rst` with Nix setup, docker compose, testing section
- Configure Poetry with explicit packages include
- Bump webpack/babel dependencies
- Only include `vitrina/` into python package to avoid polluting global namespace

Fixes: #2745


README.rst
```diff
@@ -21,55 +23,176 @@ Contributing
 .. _DDD: https://gist.github.com/zsup/9434452
 .. _GitHub Flow: https://docs.github.com/en/get-started/quickstart/github-flow
 
+System packages
+***************
+
+You can install system dependencies using your OS package manager, or use Nix
+for a reproducible development environment. Nix is optional.
+
+**Using Nix (optional, recommended):**
```
kastger, line 32: I don't think Nix should be a recommended option as most people are unfamiliar with it, including our own developers. Now almost a third of the whole README is just a setup for Nix package manager and further down the commands already assume we are using Nix (poetry run removed, the old prefix worked with zero shell state, this one requires an activated shell). I think we should fix the original version and keep it and have Nix as alternative choice.

I have recently updated this README after I had to set up everything from 0 locally and I had no issues, I am using Ubuntu with pyenv. I think we are missing a prerequisites section before starting the local setup steps, Nix does fix that, but it adds more things to maintain (flake.lock could drift after some time since no pipeline enters nix develop and recommended path becomes broken), a mostly unfamiliar package manager to users and developers and a README file that does not serve the majority very well.

Personally, I would suggest adding short "Prerequisites" sections with host requirements BEFORE poetry install - Python ^3.11, Poetry (2.4.1 known to work, we need to pin it separately), Node.js (same as Poetry, pin it and add version that's known to work), Docker, gettext (translations) and so on...

JPriev (approved):

PR #2758, opened 2026-07-30
2755 dependabot alerts

### LINKED ISSUE #2755: Resolve katalogas dependabot alerts
https://github.com/atviriduomenys/katalogas/security/dependabot

Needs to be done before every release
- [ ] If possible, create an automation in Github to create tasks from dependabot alerts.
- [x] Also, need to check what dependencies are unused and can be removed.

kastger:


webpack/package.json
JPriev, line 16: If I'm correct, this should also be ""^7.29.7". Although it's dev dependency, so leaving for you to decide.
https://github.com/atviriduomenys/katalogas/security/dependabot/136

JPriev (approved):

webpack/package.json
kastger: I will leave it as is, as the alert should resolve after merge. If it doesn't, I will fix it in a separate PR

PR #2760, opened 2026-08-03
2753 Filter dcat:servesDataset links in the EDP DCAT-AP export

### LINKED ISSUE #2753: Compare new RDF from production to EDP rdf (older)

kastger:
A data service listed the URL of every related dataset, also of non-public and deleted ones, which the European Data Portal turned into empty catalogue records.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,13 @@ https://github.com/atviriduomenys/katalogas/issues/2755
   lock file because ``django-cms`` needs it. With their own dependencies, this drops 16 packages.
 - Move ``freezegun`` and ``requests-mock`` to the dev group. The tests are their only user.
 
+https://github.com/atviriduomenys/katalogas/issues/2753
+
+- Filter ``dcat:servesDataset`` links in the EDP DCAT-AP export (`/edp/dcat-ap.rdf` and
+  `/edp/dcat-ap-restricted.rdf`). A data service listed the URL of every related dataset, also of
+  non-public and deleted ones, which the European Data Portal turned into empty catalogue records.
+  A service now links only to datasets that the export publishes.
+
 https://github.com/atviriduomenys/dvms/issues/515
 
 - Replace unsanitized ``mark_safe`` with ``format_html`` in admin display methods and form labels
--- a/tests/api/test_views.py
+++ b/tests/api/test_views.py
@@ -19,12 +19,15 @@ from vitrina.catalogs.factories import CatalogFactory
 from vitrina.classifiers.factories import CategoryFactory
 from vitrina.datasets.factories import (
     DatasetFactory,
+    DatasetRelationFactory,
+    DatasetServiceFactory,
     DatasetStructureFactory,
     DatasetGroupFactory,
     DCATResourceSubclassFactory,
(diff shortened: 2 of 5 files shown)


karina-klinkeviciute (approved):

karina-klinkeviciute:
1. /vitrina/api/helpers.py:44 — medium/high: the filter uses the union of both EDP exports instead of the export being rendered, so the reported bug still reproduces across access levels.
  _get_exported_dataset_ids() returns edp_public | edp_restricted, but edp_dcat_ap_rdf renders only Dataset.edp_public.all() and edp_dcat_ap_restricted_rdf only Dataset.edp_restricted.all()                                
  (vitrina/api/views.py:781-786). A PUBLIC data service that has a Relation.SERVICE link to a RESTRICTED dataset still emits dcat:servesDataset pointing at a dataset that /edp/dcat-ap.rdf does not describe — exactly the  
  "empty catalogue record at EDP" symptom the change is meant to remove. I confirmed this empirically: with a public service serving one restricted dataset, the public export contains /datasets/<restricted.pk>/ in a      
  servesDataset link while containing no dcat:Dataset record for it. The fix is to derive the id set from the qs actually being exported (e.g. set(qs.values_list("pk", flat=True))), and the new test should be extended    
  with a mixed-access case, since it currently parametrizes service and served dataset to the same access_rights and therefore cannot catch this.                                                                            
  2. /vitrina/api/helpers.py:49 — medium: the EDP-specific filter also silently applies to the per-dataset RDF download. get_datasets_for_rdf is shared with DatasetRDFDownloadView          
  (/vitrina/datasets/views.py:650), which renders Dataset.objects.filter(pk=...) for any user with VIEW permission. After this change, an organisation member downloading the RDF of their   
  own data service loses servesDataset links to their non-public, planned, or organisation-less datasets, even though they are allowed to see them and the download is not an EDP export. It also pays for a full            
  values_list("pk") scan over every exported dataset to render a single record. Scoping the filter to the EDP views (e.g. a flag/argument, or deriving from qs as in finding 1 — which makes this view a no-op filter) avoids
  both.                                                                                                                                                                                                                      
  3. /vitrina/api/helpers.py:106 — low: in-Python filtering keeps the existing N+1 and loads the whole catalogue into memory. The relation queryset is already hitting the DB per dataset,   
  _get_rel_dataset then dereferences relation.dataset (one more query per relation, no select_related), and _get_exported_dataset_ids() materialises every exported PK per request. Expressing the constraint in SQL —       
  dataset.related_datasets.filter(relation__name=Relation.SERVICE, dataset__in=<exported qs>).select_related("dataset") — is both cheaper and keeps the export queryset as the single source of truth.                       
  4. /vitrina/api/helpers.py:106 — low / worth confirming with the EDP requirements: a data service whose related datasets are all non-public now renders with zero dcat:servesDataset       
  children (vitrina/api/templates/vitrina/api/edp/dcat_ap_rdf.html:106-112 emits nothing). If the DCAT-AP HVD profile requires at least one servesDataset on a dcat:DataService, such services will now fail EDP validation  
  instead of producing an empty dataset record; if so, the service itself should probably be dropped from the export rather than emitted link-less.

PR #2765, opened 2026-08-05
2723 urls validator

### LINKED ISSUE #2723: Priimamų nuorodų limitai
Katalogas nepriima kai kurių vidinių nuorodų prie "AccessURL" (3. Pateiktis) bei API nuorodos is specifikacijos (EndpointURL ir EndpointDescription > "4. Paslauga"). Dėl šios priežasties negalime įvesti pateikčių ir paslaugos, kur tokios nuorodos yra netinkamos, pagal katalogą. Pvz, tai nutiko pildant VMI IMIS sistemą, kur pateiktos vidinės VMI nuorodos yra netinkamos. Nors nei viena iš šių informacijos kategorijų nėra privaloma, neturint jų, negalime užpildyti reikiamos pateikties ir paslaugos informacijos (katalogas neleidžia tokių supildymų išsaugoti).

kastger:


vitrina/validators.py
```diff
@@ -11,6 +12,22 @@
...
+    if any(character.isspace() for character in text):
+        raise ValidationError(URLValidator.message, code="invalid")
+
+    try:
+        parts = urlsplit(text)
+    except ValueError:
+        raise ValidationError(URLValidator.message, code="invalid")
+
+    if parts.scheme.lower() not in ABSOLUTE_URI_SCHEMES or not parts.netloc:
+        raise ValidationError(URLValidator.message, code="invalid")
```
karina-klinkeviciute, line 30: jei gerai suprantu, tai čia mes klaidos žinutę, kiekvieną kartą tą pačią, nepritaikytą kiekvienam atskiram atvejui?

karina-klinkeviciute (approved):

vitrina/validators.py
```diff
@@ -11,6 +12,22 @@
     message=_("Neteisingas telefono numerio formatas. Priimtini formatai: +370XXXXXXXX, 0XXXXXXXX"),
 )
 
+ABSOLUTE_URI_SCHEMES = ("http", "https", "ftp", "ftps")
+
+
+def validate_absolute_uri(value: str) -> None:
```
karina-klinkeviciute, line 18: medium. Control characters in the host are accepted and break the RDF feed.                                                                                 
  validate_absolute_uri rejects only whitespace; other C0 control characters pass because urlsplit accepts them and netloc is non-empty. _safe_uri (vitrina/api/helpers.py:24) only screens \s, and encode_uri               
  (vitrina/templatetags/util_tags.py:109) percent-encodes path/query/fragment but passes netloc through verbatim. Django's URLValidator rejected these before this change. Concrete scenario: POST a service form with       
  endpoint_url=http://ex\x01ample.com/ (Django's CharField only blocks \x00 via ProhibitNullCharactersValidator) → the value saves → /edp/dcat-ap.rdf emits rdf:resource="http://ex\x01ample.com/", and the whole catalog    
  feed is no longer well-formed XML. Verified:                                                                                                                                                                               
                                                                                                                                                                                                                             
  '<a b="http://ex\x01ample.com/p"/>'  →  XML ERROR: not well-formed (invalid token): line 1, column 15                                                                                                                      
                                                                                                                                                                                                                             
  This is the same failure class the sibling test test_edp_dcat_ap_rdf_drops_whitespace_uris guards against, and the changelog explicitly claims "the RDF rdf:resource output stays well formed". Suggest rejecting c where  
  ord(c) < 0x20 or ord(c) == 0x7f alongside the whitespace check.

vitrina/dcat/forms/distribution_forms.py
```diff
@@ -72,7 +71,6 @@ class Meta:
             "compression_format": Select2Widget,
             "packaging_format": Select2Widget,
             "checksum_algorithm": Select2Widget,
-            "download_url": URLInput,
```
karina-klinkeviciute, line 75: What the diff did                                                                                                                                                                                                          
                                                                                                                                                                                                                             
  Two edits to the same field, download_url:                                                                                                                                                                                 
   ```                                                                                                                                                                                                                          
  field_classes = {                                                                                                                                                                                                          
      "download_url": URIFormField,     # changed from forms.URLField  (intended)                                                                                                                                            
  }                                                                                                                                                                                                                          
  widgets = {                                                                                                                                                                                                                
      "download_url": URLInput,         # DELETED                      (the problem)                                                                                                                                         
  }                                                                                                                                                                                                                          
       ```                                                                                                                                                                                                                      
  How Django decides which widget to render                                                                                                                                                                                  
                                                                                                                                                                                                                             
  For a ModelForm, Django asks the model field to build the form field. Priority, highest first:                                                                                                                             
                                                                                                                                                                                                                             
  1. Meta.widgets entry — passed as an explicit argument. Always wins.                                                                                                                                                       
  2. The model field's own default — models.TextField.formfield() hardcodes "use a Textarea", and it passes that as an argument too.                                                                                         
  3. The widget = ... attribute on the form field class — only a fallback, used when nobody passed an argument.                                                                                                              
                                                                                                                                                                                                                             
  The key point: #3 only applies if #1 and #2 are both absent. It's like a default parameter value — any actual argument overrides it.   

Why the new URIFormField doesn't save you                                                                                                                                                                                  
                                                                                                                                                                                                                             
  URIFormField declares widget = TextInput, which looks like it covers the case. But that's level #3, the weakest. And DatasetDistribution.download_url is a models.TextField, so level #2 always fires and passes Textarea. 
  The TextInput on URIFormField never gets consulted.                                                                                                                                                                        
                                                                                                                                                                                                                             
  Deleting the Meta.widgets entry removed the only level-#1 override, so level #2 now wins: Textarea.                                                                                                                        
                                                                                                                                                                                                                             
  The measured results, which show the field class is irrelevant to this outcome:  

```
  ┌──────────────────┬────────────────────┬────────────────────────────┐                                                                                                                                                     
  │ form field class │ Meta.widgets entry │      rendered widget       │                                                                                                                                                     
  ├──────────────────┼────────────────────┼────────────────────────────┤                                                                                                                                                     
  │ forms.URLField   │ URLInput           │ URLInput ← before the diff │                                                                                                                                                     
  ├──────────────────┼────────────────────┼────────────────────────────┤                                                                                                                                                     
  │ forms.URLField   │ (none)             │ Textarea                   │                                                                                                                                                     
  ├──────────────────┼────────────────────┼────────────────────────────┤                                                                                                                                                     
  │ URIFormField     │ (none)             │ Textarea ← after the diff  │                                                                                                                                                     
  ├──────────────────┼────────────────────┼────────────────────────────┤                                                                                                                                                     
  │ URIFormField     │ TextInput          │ TextInput ← the fix        │                                                                                                                                                     
  └──────────────────┴────────────────────┴────────────────────────────┘   

```

  Row 2 is the important one: the old code would have broken the same way if its widgets entry had been missing. That entry was doing real work, not duplicating the field class.                                            
                                                                                                                                                                                                                             
  Why access_url is fine but download_url isn't  

```
  ┌──────────────┬──────────────────┬──────────────────────────────────────┬─────────────────────────────────────────────────────┐                                                                                           
  │    field     │    model type    │         does level #2 fire?          │                       result                        │                                                                                           
  ├──────────────┼──────────────────┼──────────────────────────────────────┼─────────────────────────────────────────────────────┤                                                                                           
  │ access_url   │ models.CharField │ No — CharField has no widget default │ falls through to URIFormField.widget → TextInput ✅ │                                                                                           
  ├──────────────┼──────────────────┼──────────────────────────────────────┼─────────────────────────────────────────────────────┤                                                                                           
  │ download_url │ models.TextField │ Yes — forces Textarea                │ Textarea ❌                                         │                                                                                           
  └──────────────┴──────────────────┴──────────────────────────────────────┴─────────────────────────────────────────────────────┘     
```

  Visible effect and fix                                                                                                                                                                                                     
                                                                                                                                                                                                                             
  The "Parsisiuntimo URL" input on the DCAT distribution form becomes a multi-line ~10-row text box instead of a single-line field.                                                                                          
                                                                                                                                                                                                                             
  One line in /home/karina/work/vssa/katalogas/vitrina/dcat/forms/distribution_forms.py:65:                                                                                                                                  
                              
```                                                                                                                                                                                               
  widgets = {                                                                                                                                                                                                                
      ...                                                                                                                                                                                                                    
      "download_url": forms.TextInput,                                                                                                                                                                                       
  }   

```

vitrina/validators.py
kastger: Yes, naudojam Django `URLValidator.messsage` kuriam jau yra vertimas ir žinutę sutampa su ankstesne.

vitrina/validators.py
kastger: Fixed

vitrina/dcat/forms/distribution_forms.py
kastger: Fixed and added tests to cover these scenarios.

PR #2768, opened 2026-08-07
513 DVMS XSS Mermaid XSS

JPriev:


vitrina/structure/services.py
```diff
@@ -2588,6 +2599,12 @@ def get_allowed_visibilities(
     return allowed_visibilities
 
 
+def _escape_mermaid(value: str) -> str:
+    if not value:
+        return value
+    return value.replace("\r", " ").replace("\n", " ").translate(_MERMAID_UNSAFE_CHARS)
```
kastger: Image

This is the same function as in `spinta`. Can we use the same names and maybe add a comment that its the same as `spinta` so if we see any issues with it we know to fix it in two places.

vitrina/structure/services.py
```diff
@@ -2588,6 +2599,12 @@ def get_allowed_visibilities(
     return allowed_visibilities
 
 
+def _escape_mermaid(value: str) -> str:
```
kastger: Type hints should be `str | None` for input and ouput

vitrina/structure/services.py
```diff
@@ -64,6 +64,17 @@
 
 MetadataCache = Dict[str, Metadata]
 
+_MERMAID_UNSAFE_CHARS = str.maketrans(
```
kastger: Rename to `_MMD_ENCODE` same as in Spinta maybe? Since its the same value and usage

vitrina/structure/services.py
```diff
@@ -64,6 +64,17 @@
...
+    {
+        "`": "#96;",
+        '"': "#34;",
+        "{": "#123;",
+        "}": "#125;",
+        "<": "#60;",
+        ">": "#62;",
```
kastger: Maybe we should add `"\r"`, `"\n"` to the list too as `" "`

vitrina/structure/services.py
kastger: Move the this closer to `_escape_mermaid` so either go down to 2602 line or move the function here

vitrina/structure/templates/vitrina/structure/uml_canvas.html
```diff
@@ -128,8 +128,16 @@
...
-        sceneElement.innerHTML = svg;
+
+        let svg;
+        try {
+            ({ svg } = await mermaid.render('diagram-1', mermaidCode));
+            sceneElement.innerHTML = svg;
+        } catch (error) {
+            console.error('Nepavyko atvaizduoti UML diagramos:', error);
+            sceneElement.innerHTML =
+                '<div class="uml-loader"><p>{% translate "Įvyko klaida atvaizduojant diagramą." %}</p></div>';
```
kastger: Since this is translator-editable, one apostrophe can become a problem, we need to `escapejs`, same as in line 159 (150 prior these changes)

kastger: Image

kastger:
```
<div class="uml-loader"><p>{% translate "Įvyko klaida atvaizduojant diagramą." as uml_error %}{{ uml_error|escapejs }}</p></div>'
```

vitrina/structure/templates/vitrina/structure/uml_canvas.html
```diff
@@ -119,7 +119,7 @@
 
     mermaid.initialize({
         startOnLoad: false,
-        securityLevel: 'loose',
+        securityLevel: 'antiscript',
```
kastger, line 122: Would be great to add regression test regarding the `securityLevel`, something like:
```
assert response.status_code == 200
assert "securityLevel: 'antiscript'" in response.text
assert "securityLevel: 'loose'" not in response.text
```

In `TestStructureUMLviews.test_uml_view`

PR #2769, opened 2026-08-07
[DAS-643] anonymize.py: fix wrong target tables, add --keep-public-content

SauliusPetraitis:
JIRA: [DAS-643](https://itpagalba-vssa.atlassian.net/browse/DAS-643)

Found while preparing the django-cms 5 dry run ([DAS-428](https://itpagalba-vssa.atlassian.net/browse/DAS-428)), but everything here is about the script as it stands today, on django-cms 3.

## 1. Two functions wrote to the wrong table

`_anonymize_organization` operated on `djangocms_blog_post_translation` instead of `organization` — a copy-paste slip from `3536317d` ("185: new tables in anonymization script"). **Organisation e-mails, phone numbers and addresses were therefore never anonymized.**

Measured before/after on a real dump: `519/525 -> 525/525`. In the `adp-prod-2026-03-19` dump six organisations still carried real domains (`savivaldybe.lt`, `kgriniausligonine.lt`, `detonas.eu`, `qwer.lt`, `testing.com`, `zua.lt`).

`_anonymize_adp_cms_page` operated on `news_item`, so `adp_cms_page` was never covered and `news_item` was anonymized twice.

Worth noting: because `dataset` creates missing columns instead of failing, this went unnoticed. `djangocms_blog_post_translation` still carries stray `address`, `email` and `phone` columns that the buggy function created.

## 2. `faker` was not declared

The script imports it, but only `dataset` was in `pyproject.toml`. On a clean install it fails with `ModuleNotFoundError` before doing anything. Added to the dev group — anonymization runs from a dev/ops environment, not from the production image.

## 3. New `--keep-public-content`

Keeps **published** news and CMS page texts, titles and slugs readable, so that content can actually be reviewed. Without it every article becomes `example` with body `<p>example</p>` and a review proves nothing. Default behaviour is unchanged.

Personal data is still anonymized everywhere:

- drafts, non-public and soft-deleted rows are anonymized as usual — only what the site already shows publicly is spared (`djangocms_blog_post.publish`, `news_item.is_public` and not `deleted`, `adp_cms_page.published`)
- the byline is removed from its own columns (`news_item.author_name`, `news_item.author`) **and from inside the article text**, where `scripts/migrate_news.py` writes it as `<p>Autorius: NAME</p>` — keeping the text would otherwise keep the name

`news_item.author` is a legacy **bytea** column even though the model declares `TextField`, so it needs bytes; writing a `str` fails with `TypeError: can't escape str to binary`. It is empty in the current production copy, so this is prevention rather than a fix for existing data.

## Verification

Sentinels on a real copy, `--keep-public-content`:

| | before | after |
|---|---|---|
| published article text | `<p>Autorius: Jonas Jonaitis</p><p>TEKSTAS-PUB</p>` | `<p>Autorius: example, example</p><p>TEKSTAS-PUB</p>` |
| draft article | `ANTRASTE-DRAFT` | `example` |
| public news item | `Ona Onaite` + name in body | `example, example` in both |
| non-public news item | `NEWS-PRIVATE / Slaptas Vardas` | `example / example, example` |
| published CMS page | `PAGE-PUB` | `PAGE-PUB` |
| real names left in text | — | **0** |

Organisation fix measured separately: `519/525 -> 525/525`.

## Scope note

An earlier revision also made the script pick between `djangocms_blog_post_translation` and `djangocms_stories_postcontent` at run time. Dropped after discussing it with @kastger: the script should not carry both the cms 3 and the cms 5 shape, because once the site is on cms 5 the cms 3 branch is dead code nobody removes. Adapting the script to `djangocms_stories` is recorded as part of the django-cms 5 upgrade work ([DAS-647](https://itpagalba-vssa.atlassian.net/browse/DAS-647)) instead.

## Usage

```sh
python scripts/anonymize.py <db-uri> --yes --keep-public-content
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scripts/anonymize.py
```diff
@@ -85,8 +101,23 @@ def main(
...
 
 
+def _table_for(db: Database, name: str) -> str:
+    """Loginį vardą paverčia realiu lentelės vardu.
+
+    Po django-cms 5 atnaujinimo `djangocms_blog_*` lentelių nebelieka — turinys
+    persikelia į `djangocms_stories_*`. Laukų vardai sutampa, tad skiriasi tik lentelė.
+    Skriptas turi veikti IR prieš atnaujinimą, IR po jo: kitaip pirmas dump'as po
+    atnaujinimo liktų be veikiančio anonimizavimo.
+    """
```
kastger: We usually write the docstrings in English

scripts/anonymize.py
```diff
@@ -85,8 +101,23 @@ def main(
...
+    if name == "post_content":
+        if "djangocms_stories_postcontent" in db.tables:
+            return "djangocms_stories_postcontent"
+        return "djangocms_blog_post_translation"
+    return name
```
kastger: Function uses hardcoded values and use same string twice. Better to have a module level variable such as `_POST_CONTENT_TABLES = ("djangocms_stories_postcontent", "djangocms_blog_post_translation")`

Also, functions fallback is silent, if neither table appears in the database, the helper still returns the legacy name. The anonymization will silently do nothing, it will look like it was finished successfully, but no actual anonymization would be done
```
def _table_for(db: Database, name: str) -> str:
    if name != "post_content":
        return name
    for table in _POST_CONTENT_TABLES:
        if table in db.tables:
            return table
    raise RuntimeError(f"No post content table found. Looked for: {_POST_CONTENT_TABLES}")
```

pyproject.toml
```diff
@@ -79,6 +76,7 @@ python-magic = "^0.4.27"
 django-csp = "^4.0"
 
 [tool.poetry.group.dev.dependencies]
+faker = "*"
```
kastger: Dependency is not needed, we have `factory-boy` which requires `faker`. The script failed with ``ModuleNotFoundError`` on a clean install likely because you installed without the dev group.

scripts/anonymize.py
```diff
@@ -10,9 +10,22 @@
...
+
+
 def main(
     uri: str = Argument(..., help=("Database URI (postgresql://user:pass@host:port/db)")),
     yes: bool = Option(False, help="Do not ask anything, just do it"),
+    keep_public_content: bool = Option(
+        False,
+        help=(
+            "Keep already published editorial content readable: news and CMS page texts, "
+            "titles and slugs. Personal data is still anonymized everywhere, including the "
```
kastger: `...Personal data is still anonymized everywhere` is not true. Text columns such as `body`, `abstract` and `post_text` can still hold personal data, and they are not scanned.

scripts/anonymize.py
```diff
@@ -21,13 +34,16 @@ def main(
     restore it. So before using this command, make sure to pass correct database
     connection string.
     """
+    global _KEEP_PUBLIC_CONTENT
+    _KEEP_PUBLIC_CONTENT = keep_public_content
+
```
kastger, line 44: This is not needed, just use `keep_public_content`. Add instead, after the `tables` tuple and before the `total`:

```
_PUBLIC_CONTENT_TABLES = ("post_content", "adp_cms_page")
if keep_public_content:
    echo(
        "WARNING! --keep-public-content keeps published text as is. That text is not scanned, "
        "so it can still hold names, e-mail addresses or phone numbers."
    )
    tables = tuple(t for t in tables if t not in _PUBLIC_CONTENT_TABLES)
```

kastger:
More to review, but need to resolve merge conflicts before moving on

SauliusPetraitis:
Both points addressed:

**Merge conflicts** — resolved. `devel` had moved 11 commits ahead (the Nix dev env PR #2757 among them), which is where the `pyproject.toml` / `poetry.lock` conflicts came from. `poetry.lock` regenerated from the current `devel` rather than hand-merged, and the `CHANGES.rst` entry for #2762 kept alongside this one. PR is mergeable again.

**Docstrings in English** — fixed, and not only the one you flagged: `_table_for`, `_anonymize_post_content`, `_anonymize_news_item` and `_anonymize_adp_cms_page` all had Lithuanian comments. All translated in `85a39ae3`, kept as a separate commit so it is easy to see what changed since your review.

Note `scripts/` is outside the ruff `include` (`vitrina/*.py`, `tests/*.py`), so nothing here is linted automatically.

Ready for the rest of the review.

SauliusPetraitis:
Scope reduced as agreed: the run-time schema detection is out (`d08fb056`), so this PR now only touches the script as it works today on cms 3 — the two wrong-table fixes, the missing `faker` dependency and `--keep-public-content`.

Adapting the script to `djangocms_stories` is recorded as part of the django-cms 5 upgrade work instead, so it does not get forgotten there.

SauliusPetraitis:
Ran a critical review over this branch (Fable) with the brief "a defect here means personal data ends up in a copy people treat as safe". It found three real problems in `--keep-public-content`, all now fixed and verified.

**1. The byline was kept in plain text.** `scripts/migrate_news.py:32-34` prepends `<p>Autorius: NAME</p>` to `body` and `summary`, and those become `post_text` / `abstract` on the blog post. The flag skipped that table wholesale, so the name survived — in exactly the mode meant for sharing dumps, and directly against what the flag's own help promised. Now the text is kept but the byline inside it is replaced.

**2. The flag spared everything, not only published rows.** The help said "already published", the code skipped drafts, non-public and soft-deleted rows too. Now it only spares rows the site actually shows (`djangocms_blog_post.publish`, `news_item.is_public` and not `deleted`, `adp_cms_page.published`); everything else is anonymized as before.

**3. `news_item.author` was never anonymized** in either mode. Note it is a legacy **bytea** column even though the model declares `TextField`, so it needs bytes — writing a `str` fails with `TypeError: can't escape str to binary`. In the current production copy the column is empty, so this is prevention rather than a fix for existing data.

Verified with sentinels on a real copy:

| | before | after |
|---|---|---|
| published article text | `<p>Autorius: Jonas Jonaitis</p><p>TEKSTAS-PUB</p>` | `<p>Autorius: example, example</p><p>TEKSTAS-PUB</p>` |
| draft article | `ANTRASTE-DRAFT` | `example` |
| public news item | `Ona Onaite` + name in body | `example, example` in both |
| non-public news item | `NEWS-PRIVATE / Slaptas Vardas` | `example / example, example` |
| published CMS page | `PAGE-PUB` | `PAGE-PUB` |
| real names left in text | — | **0** |

The review also flagged the branch as carrying unrelated changes (a `spinta` major bump). That one was a false alarm — it compared against a stale local `devel`; `git diff origin/devel...das-643-anonymize -- pyproject.toml` shows only `+faker = "*"`.

SauliusPetraitis:
All addressed in `8f816411`. Thanks — several of these were real.

**@kastger**

- *docstrings in English* — done earlier (`85a39ae3`), and not only the one line you marked: `_table_for`, `_anonymize_post_content`, `_anonymize_news_item` and `_anonymize_adp_cms_page` all had Lithuanian comments.
- *`_table_for` hardcoded strings / silent fallback* — the whole helper is gone. After we agreed the script should not carry both the cms 3 and cms 5 shape, the run-time table lookup was dropped in `d08fb056`. Your point about the silent fallback was right though, and it is worth remembering for the cms 5 upgrade PR, where this logic will reappear.
- *`faker` not needed* — correct, `factory-boy` requires `Faker >= 0.7.0` and it is already in the dev group. My `ModuleNotFoundError` came from running inside an image built with `poetry install --only main`. Declaration removed. Worth noting separately that this means `scripts/anonymize.py` cannot be run from a production image at all — it needs the dev group.
- *"Personal data is still anonymized everywhere" is not true* — agreed, and this was the worst line in the PR. Free text in `body`, `abstract` and `post_text` is never scanned; an editor can have typed a name or a phone number into an article. Reworded, and the command now prints a warning on start when the flag is used.
- *drop the module-level flag, filter the `tables` tuple instead* — I did not take this one, and want to explain why rather than ignore it. Removing `post_content` and `adp_cms_page` from the tuple would skip those tables entirely, which is what the first revision did and exactly what leaked: `scripts/migrate_news.py` writes `<p>Autorius: NAME</p>` into the article text, so skipping the table keeps the byline. The rows still have to be walked — to scrub the byline and to tell published rows from drafts. The warning part of your suggestion I did take.

**@Copilot**

- *`publish` alone is not "currently public"* — correct. `djangocms-blog`'s `published()` also requires `date_published <= now` and `date_published_end` to be unset or in the future. A scheduled post would have been spared. Now uses the full predicate. There are no such rows in the current copy, so this is prevention.
- *`author` bytes break the canonical `text` schema* — correct, and better than what I had. `vitrina/cms/migrations/0001_initial.py:125` declares it as a nullable `TextField`; production has it as `bytea`. `None` removes the byline in both. Changed in both branches.
- *published but soft-deleted page was spared* — correct, `deleted` is now part of the condition.

SauliusPetraitis:
Re-ran the sentinel test on a real copy after the review fixes, extended to cover the cases you both raised. All pass.

| case | before | after | expected |
|---|---|---|---|
| published post | `ANTRASTE-6` | `ANTRASTE-6` | kept |
| draft post | `ANTRASTE-1` | `example` | anonymized |
| **scheduled** (`date_published` +30d) | `ANTRASTE-7` | `example` | anonymized |
| **expired** (`date_published_end` −30d) | `ANTRASTE-8` | `example` | anonymized |
| public news item | `NEWS-PUB / Ona Onaite` | `NEWS-PUB / example, example / author=NULL` | text kept, byline gone |
| non-public news item | `NEWS-PRIVATE` | `example` | anonymized |
| published page | `PAGE-PUB` | `PAGE-PUB` | kept |
| **published but deleted page** | `PAGE-DELETED` | `example` | anonymized |
| real name left in article text | — | **0** | none |
| warning printed on start | — | yes | shown |

The scheduled, expired and deleted rows are the three cases that would have leaked before; they are now anonymized like any other non-public row. `author` comes out as `NULL`, which is what works on both the production `bytea` column and the canonical `text` one.

The run also installed with `poetry install --only dev` and no explicit `faker` declaration, so that confirms your point — `factory-boy` brings it in.

kastger (changes requested):
What are these removals in poetry.lock ? How are they related to these changes?

SauliusPetraitis:
Note for anyone reading the thread: the three Copilot comments above were made at 16:44 UTC against `7ac828ac`, and all three were fixed 21 minutes later in `8f816411` — the scheduled/expired `date_published` predicate, `author = None` instead of bytes, and the `deleted` flag on published pages. They are outdated rather than open. Verified with sentinels covering exactly those three cases (see the table two comments up).

kastger (approved):

PR #2779, opened 2026-08-13
1825 es remove

### LINKED ISSUE #1825: Upgrade Elasticsearch to 9.x (or at least 8.x)
The upgrade of the Elasticsearch version is currently blocked by the `django-haystack` library, which the catalog uses for Elasticsearch searching. The existing version of `django-haystack` does not support Elasticsearch 8; therefore, to upgrade Elasticsearch, we must first deprecate and remove `django-haystack.`

Currently, `django-haystack` restricts and requires Elasticsearch 7.8.1 (which is EOL):

https://github.com/django-haystack/django-haystack/issues/1980

```toml
[project.optional-dependencies]
elasticsearch = [
  "elasticsearch<8,>=5",
]
```

https://github.com/django-haystack/django-haystack/blob/39d0dc1f5a730bb916b684ed1945eda0017b9783/pyproject.toml#L45-L48

## Other Related Issues

- https://github.com/atviriduomenys/katalogas/issues/2747
- https://github.com/atviriduomenys/katalogas/issues/2746
- https://github.com/atviriduomenys/katalogas/issues/2261
- https://github.com/atviriduomenys/katalogas/issues/2257
- https://github.com/atviriduomenys/katalogas/issues/2298
- https://github.com/atviriduomenys/katalogas/issues/1925

## Proposed Solution

Completely remove both `django-haystack` and Elasticsearch, and migrate fully to PostgreSQL:

- **Materialized Views**: Instead of synchronizing Elasticsearch indexes, we can utilize PostgreSQL Materialized Views. This eliminates the need for expensive and complex synchronization between two separate databases.

- **Faceted Search & RAG**: Use `django-paradedb` to handle advanced search features.

With this change operational simplification would be significant:

- No separate ES service (one fewer Docker container)
- No index synchronization issues
- No 40,972 HTTP calls per test suite
- No django-haystack (unmaintained since 2013)
- No ES 7.8.1 (EOL)
- Search results become just SQL queries — debuggable with EXPLAIN ANALYZE
- The dataset is small enough that PG will outperform ES for this workload

## Sources

- [Haystack -> ElasticSearch](https://django-haystack.readthedocs.io/en/master/backend_support.html#id2) - Requires: elasticsearch-py 1.x, 2.x, 5.X, or 7.X.
- [ElasticSearch version support matrix](https://www.elastic.co/support/matrix)
- [ElasticSearch release notes](https://www.elastic.co/docs/release-notes/elasticsearch)

## Related

- https://github.com/atviriduomenys/katalogas/issues/1470

kastger:
Removes Elasticsearch and `django-haystack`; `katalogas` now searches PostgreSQL through the new `vitrina.search` app, which keeps one document row and a set of facet rows per searchable model and refreshes them from a `post_save` receiver inside the same transaction.

Most of the changed files are `@pytest.mark.haystack` removals, `i.object.X` -> `i.X` rewrites across three list templates, the deleted index modules, and config. The parts worth real attention are `vitrina/search/` and `filter_out_non_public_datasets_for_user` in `vitrina/datasets/services.py`, which is the gate deciding who sees non-public datasets. Five things are intentional rather than bugs: the request list now sorts by title (it previously sorted by fields the request index never held), every list order ends with `pk` so pages cannot repeat or drop rows, search words of one or two characters match a word start while longer ones match any position (Elasticsearch matched only word starts, so results widen slightly), the `search/indexes/**/*_text.txt` templates stay because `indexing.py` renders them, and there is no `post_delete` receiver because the foreign keys cascade.


vitrina/search/signals.py
```diff
@@ -0,0 +1,56 @@
...
+    fresh = copy(instance)
+    for name in _STALE_CACHES:
+        fresh.__dict__.pop(name, None)
+    fresh._prefetched_objects_cache = {}
+    return fresh
+
+
+def _reindex_dataset(instance):
+    from vitrina.requests.models import RequestObject
+    from vitrina.search.indexing import index_dataset, index_request
```
JPriev: Is it mandatory to have these imports inside function?

vitrina/search/indexing.py
```diff
@@ -0,0 +1,250 @@
...
+    "access_rights": lambda dataset: [dataset.access_rights],
+    "subclass": lambda dataset: [dataset.subclass.name if dataset.subclass_id else None],
+}
+
+
+def _request_categories(request):
+    categories = []
+    if request.dataset_id and request.dataset.category:
+        for category in request.dataset.category.all():
+            categories = [cat.pk for cat in category.get_ancestors() if cat.dataset_set.exists()]
```
JPriev: Any particural reason why this is not `.extended` like `_dataset_categories` above? It might be out of scope, but it's worth mentioning
`_dataset_categories` also were reassigning `categories` values prior to this fix #1102

vitrina/search/queries.py
```diff
@@ -0,0 +1,91 @@
...
+    if not found:
+        return []
+
+    result = []
+    year, month = min(found).year, min(found).month
+    last = max(found)
+    while (year, month) <= (last.year, last.month):
+        bucket = date(year, month, 1)
+        result.append((bucket, found.get(bucket, 0)))
+        year, month = (year + 1, 1) if month == 12 else (year, month + 1)
```
JPriev, line 92: Before it had set `start_date` and `end_date`

```python
# vitrina/datasets/views.py:188-195
date_facet_fields = [
    {
        "field": "published",
        "start_date": date(2019, 1, 1),
        "end_date": date.today(),
        "gap_by": "month",
    },
]
```

And now it doesn't. Through `/admin` it is possible to set published date manually. So in **edge case** if admin would make a typo and set it to 2099 instead of 2019, it would fill the gaps with zeros for ~80 years.

It's just an observation to think about what problems we could face from it

vitrina/search/indexing.py
```diff
@@ -0,0 +1,250 @@
...
+def _facet_rows(obj, facets):
+    for field, get_values in facets.items():
+        for value in get_values(obj) or ():
+            if value is None or value == "":
+                continue
+            yield field, str(value)
+
+
+def _render_text(obj, template):
+    return render_to_string(template, {"object": obj})
```
JPriev: This autoescapes, so a title containing `"` is indexed as `&quot;` (same for
`&`, `'`, `<`, `>`). `text_query` then matches the raw keyword, so the query
never matches its own indexed text — `?q=Vilniaus` finds the dataset,
`?q="vandenys"` returns nothing.

`html.unescape` around it fixes it, but existing rows keep the entities, so it
needs a `rebuild_search` too. Worth a test — `tests/search/` has no
punctuation case.

Image

kastger: Fixed and added regression tests

vitrina/search/queries.py
kastger: Fixed

vitrina/search/signals.py
kastger: Not mandatory, `singals.py` is only imported from `ready()`, so the registry is complete and there is no import cycle. Moved.

vitrina/search/indexing.py
kastger: Reused code from the old `prepare_category`. #1102 fixed only the dataset side and left this one out. Fixed now.

vitrina/search/indexing.py
```diff
@@ -0,0 +1,251 @@
...
+}
+
+
+def _request_categories(request):
+    categories = []
+    if request.dataset_id and request.dataset.category:
+        for category in request.dataset.category.all():
+            categories.extend([cat.pk for cat in category.get_ancestors() if cat.dataset_set.exists()])
+            categories.append(category.pk)
+    return categories
```
JPriev: `_request_categories` and `_dataset_categories` functions have identical bodies apart from the source queryset - worth extracting a shared helper.

JPriev (approved):

PR #2781, opened 2026-08-19
2746 dataset cpu cost

### LINKED ISSUE #2746: `/datasets/` (and requests/orgs) list pages cost ~1 CPU-second per render
## Summary

The three search-result list views (`datasets`, `requests`, `orgs`) render each result by dereferencing `i.object` in the template. Each dereference lazily re-loads the full model instance from Postgres (one round-trip per row) and then fires a series of **unbatched per-object queries** - icon/category walk, organization FK, maturity level, groups, formats, likes, hit count. At 20 results per page this is on the order of ~10 queries per result plus the per-row object load, plus up to ~50 for the tags facet and the CMS menu - roughly 150-250 queries and a large amount of Python-side query construction per page load.

The endpoint costs roughly **1 CPU-second per render**, almost all of it in the app layer, not the database. This makes the list pages the dominant CPU cost on the site and its point of failure under load.

## Why now (impact)

Two production incidents (2026-07-19/20 and 2026-07-27) were distributed scrapes of `/datasets/` that saturated all gunicorn workers and caused sustained 502 responses.

Measurements from the 2026-07-27 event:

- `vitrina` (app) container at **~1200% CPU** while `postgres` sat at **264%** - the bottleneck is Python building/rendering querysets, not SQL execution.
- `py-spy` stacks (2026-07-19) showed every worker inside `DatasetListView`, CPU-bound in Django's query compiler.

The code had not changed; traffic simply crossed a line an always-expensive endpoint always had. **~30 req/s of legitimate traffic would produce the same outage** - this is not only a bot problem. The site is currently held up by an nginx tourniquet (global rate limit on `/datasets/` + datacenter IP bans), which is a stopgap, not a fix.

## Evidence: the code

All references on `devel` @ `efeae5ae`.

### Per-result `i.object` dereference (the core issue)

`vitrina/datasets/templates/vitrina/datasets/list.html` renders each of 20 results via `i.object`, which triggers haystack's lazy DB fetch (one round-trip per row) plus:

| Line | Call | Cost |
|---|---|---|
| 183 | `{% if i.object %}` | lazy load of the full `Dataset` (per row) |
| 186 | `i.object.get_icon` | walks `category.all()` + `get_root()` per category, then a `Category` query (`models.py:1148`) |
| 222 | `i.object.organization.title` / `.get_absolute_url` | organization FK fetch |
| 240 | `i.object.get_level` | `metadata.first()` query (`models.py:1182`) |
| 256 | `i.object.description\|markdown` | per-result markdown render (CPU) |
| 264 | `i.object.get_all_groups` | 2 queries: category→group ids, then `DatasetGroup` filter (`models.py:858`) |
| 267 | `i.object.distinct_formats` | `is_part_of_dataservice()` EXISTS + up to 3× `Format.objects.get(title=...)` (`models.py:946`) |
| 277 | `{% get_hit_count for i.object %}` | django-hitcount generic-FK query |
| 284 | `i.object.get_likes` | `ContentType.get_for_model` + `Like.objects.filter().count()` (`models.py:1217`) |

`get_download_count` (`models.py:1226`) is already cheap - it reads the denormalized `download_count` field.

Same pattern in the other two list templates:

- `vitrina/requests/templates/vitrina/requests/list.html` - note `i.object.requestassignment_set.all` (lines 145-146), one query per result.
- `vitrina/orgs/templates/vitrina/orgs/list.html:145` - `i.object.get_absolute_url` still forces the per-row object fetch even though display otherwise uses `i.title`.

### Two shared contributors (fire on every list render)

1. **Tags facet - one query per value.** `vitrina/helpers.py:150` (`Filter.items`,  `display_method` branch) resolves each facet value individually.

2. **Uncached CMS menu.** `vitrina/templatetags/navigation_tags.py:8` (`show_menu`) runs 2 base queries plus **one subquery per top-level nav page**, uncached, on every render (`navigation_tags.py:11`).

### Why rendering from the search index is safe

Permission filtering is applied as an Elasticsearch filter, not per-object in the template (`get_datasets_for_user` → `filter_out_non_public_datasets_for_user`, `vitrina/datasets/services.py`), so reading display fields from the index does not weaken access control. Most display data is already indexed (`DatasetIndex`, `vitrina/datasets/search_indexes.py`), and list URLs are pk-based.

### 1. Load test (committed, reusable)

A locust (https://github.com/locustio/locust) script committed to the repo (`loadtest/locustfile.py` + README, `locust` in a poetry dev group) that:

- Discovers valid facet IDs from the app on startup (no hand-maintained URL list) and assembles random `/datasets?selected_facets=…` combinations, reproducing the scrape pattern.
- Sweeps facet depth (0 / 1-3 / 15-20) - different depths hit different cost paths (per-result render vs facet aggregation).
- Ramps request rate to find the saturation knee (req/s vs p95 latency vs error rate).

Purpose: (a) establish the real baseline the "~30 req/s" estimate stands in for, (b) prove the fix's before/after multiple, (c) guard against regression.

### 2. The fix: render list pages from the index

Read display data from the search index; batch anything that must stay live.

- Static-ish display fields (title, dates, icon, organisation, level, groups, formats) → stored index fields.
- Live counters (likes, hits) → batched per-page into `{pk: count}` dicts, one grouped query each over the page's pks. `download_count` is already a field. 
-  Reference lookups (group/format titles) → `in_bulk()` once per page.
- Extend bulk fetch to the `display_method` branch of `Filter.items` (tags facet).
- Cache `show_menu`.

## Related

- https://github.com/atviriduomenys/katalogas/issues/1825

kastger:
| URL | devel | branch, before the icon fix | branch now | total gain |
| --- | --- | --- | --- | --- |
| `/` (landing) | 0.072 s / 101 q | 0.031 s / 44 q | 0.034 s / 45 q | 2.1x CPU, 2.2x queries |
| `/datasets/` | 0.228 s / 229 q | 0.117 s / 51 q | 0.109 s / 28 q | 2.1x CPU, 8.2x queries |
| `/datasets/?page=5` | 0.255 s / 299 q | 0.129 s / 74 q | 0.103 s / 29 q | 2.5x CPU, 10.3x queries |
| `/datasets/` + tag facet | 0.247 s / 313 q | 0.118 s / 79 q | 0.087 s / 29 q | 2.8x CPU, 10.8x queries |
| `/datasets/?q=duomenys` | 0.252 s / 298 q | 0.121 s / 73 q | 0.096 s / 29 q | 2.6x CPU, 10.3x queries |
| `/requests/submitted/` | 0.051 s / 66 q | 0.022 s / 12 q | 0.023 s / 12 q | 2.2x CPU, 5.5x queries |
| `/organizations/` | 0.133 s / 577 q | 0.006 s / 8 q | 0.006 s / 8 q | 22x CPU, 72x queries |

Measured on a local anonymized dump of 2,891 indexed datasets, `DEBUG=False`, median of 10 renders: `/datasets/` drops from 0.228 to 0.109 CPU seconds and from 229 to 28 queries, and the query count is now flat in both the row count and the row content. `/organizations/` drops from 0.133 to 0.006 CPU seconds and from 577 to 8 queries.


vitrina/datasets/services.py
```diff
@@ -776,3 +779,49 @@ def _build_dataset_url(self) -> str:
...
+        dataset.like_count = likes.get(dataset.pk, 0)
+        dataset.hit_count = hits.get(dataset.pk, 0)
+        dataset.display_groups = [group for pk, group in groups.items() if pk in selected]
+
+        formats = dataset.formats
+        if dataset.serves_dataservice_formats():
+            if dataservice_formats is None:
+                dataservice_formats = list(Format.objects.filter(title__in=DATASERVICE_FORMAT_TITLES))
+            formats = formats + dataservice_formats
+        dataset.display_formats = sorted(set(formats), key=lambda fmt: fmt.title)
```
sirex: This looks very similar to `Dataset.distinct_formats` method. I would suggest to add  a `Dataset.get_available_formats()` method and place it near the `Dataset.distinct_formats` method and extract duplicate logic if possible.

vitrina/datasets/services.py
```diff
@@ -776,3 +779,49 @@ def _build_dataset_url(self) -> str:
...
+    }
+    roots = []
+    if any(root_paths.values()):
+        roots = list(Category.objects.filter(depth=1).order_by("title").values_list("path", "icon"))
+
+    dataservice_formats = None
+
+    for dataset in datasets:
+        selected = group_pks_by_dataset[dataset.pk]
+        dataset.icon = next((icon for path, icon in roots if path in root_paths[dataset.pk]), None)
```
sirex: Looks similar to `Dataset.get_icon()` method. I would suggest to use existing or add a new method, instead of introducing transient fields to the `Dataset` instance.

vitrina/datasets/services.py
```diff
@@ -776,3 +779,49 @@ def _build_dataset_url(self) -> str:
...
+        dataset.like_count = likes.get(dataset.pk, 0)
```
sirex: Looks similar to `Dataset.get_likes()` method. I would suggest to use existing or add a new method, instead of introducing transient fields to the `Dataset` instance.

vitrina/datasets/services.py
```diff
@@ -776,3 +779,49 @@ def _build_dataset_url(self) -> str:
...
+        dataset.hit_count = hits.get(dataset.pk, 0)
```
sirex: I would suggest to add a new method `Dataset.get_git_count()`, instead of introducing transient fields to the `Dataset` instance.

vitrina/datasets/services.py
```diff
@@ -776,3 +779,49 @@ def _build_dataset_url(self) -> str:
...
+        dataset.display_groups = [group for pk, group in groups.items() if pk in selected]
```
sirex: I would suggest to build a dict as template context, instead of introducing transient fields to the `Dataset` instance.

tests/datasets/test_list_query_count.py
```diff
@@ -0,0 +1,133 @@
...
+    with CaptureQueriesContext(connection) as few:
+        app.get(reverse("dataset-list"))
+
+    for _ in range(12):
+        DatasetFactory(category=[category])
+    with CaptureQueriesContext(connection) as many:
+        app.get(reverse("dataset-list"))
+
+    growth = len(many) - len(few)
+    assert growth <= 2, f"{growth} extra queries for 12 extra categorised results ({len(few)} -> {len(many)})"
```
sirex: All tests looks the same, maybe they could be parametrized?

tests/datasets/test_list_rendering.py
```diff
@@ -0,0 +1,86 @@
...
+    HitCount.objects.create(
+        content_type=ContentType.objects.get_for_model(Dataset),
+        object_pk=dataset.pk,
+        hits=4471,
+    )
+
+    response = app.get(reverse("dataset-list"))
+
+    assert response.context["object_list"][0].hit_count == 4471
+    assert "4471" in response.text
```
sirex, line 28: This looks like a unit test, not sure if it is worth to call a view, instead of a function directly. Same thing, with other tests.

I would keep most of the tests, unit-test style, fast and testing functions directly, and only add few functional tests, that call views and test overall logic.

Unit tests should be fast and easy to maintain.

tests/datasets/test_list_rendering.py
```diff
@@ -0,0 +1,86 @@
...
+    assert response.context["object_list"][0].icon == dataset.get_icon() == ""
+
+
+@pytest.mark.django_db
+def test_dataset_list_icon_is_none_without_a_category(app: DjangoTestApp):
+    dataset = DatasetFactory()
+
+    response = app.get(reverse("dataset-list"))
+
+    assert response.context["object_list"][0].icon == dataset.get_icon() is None
```
sirex: It is quote rare to see chained comparison.. 😲

tests/orgs/test_list_query_count.py
```diff
@@ -0,0 +1,36 @@
...
+def test_organization_list_query_count_does_not_grow_with_jurisdiction_count(app: DjangoTestApp):
+    _organizations_in_new_jurisdictions(2)
+    app.get(reverse("organization-list"))
+
+    with CaptureQueriesContext(connection) as few:
+        app.get(reverse("organization-list"))
+
+    _organizations_in_new_jurisdictions(10)
+    with CaptureQueriesContext(connection) as many:
+        app.get(reverse("organization-list"))
```
sirex, line 33: The `vitrina/orgs/templates/vitrina/orgs/list.html` still dereferences `i.get_absolute_url`.

sirex (approved):
The performance books is good, bet I don't like the `annotate_dataset_list_rows` function. Looks like a hack, adds transient fields on Dataset instances, looks like duplicates some logic.

It looks, that we need the [CQRS](https://martinfowler.com/bliki/CQRS.html) + [Event Sourcing](https://github.com/pyeventsourcing/eventsourcing).

This way, we can keep the clean Conceptual data model, and derive large denormalized read tables, for specific UI parts.

This way, we will have clean logic in one places, and very fast reads.

Of course, this would come with some added complexity.

But it would fix multiple performance related issues, Elastic Search indexing, DSA related issues, access rights management issues, etc.

Other comments are added inline.

tests/datasets/test_list_rendering.py
kastger: The query-count tests measure the view, so they have to call it. The value tests also hold the name the template reads: if the view and the template disagree, Django renders an empty string and a test on the function alone still passes. These failures are quiet - the hit count once rendered 0 on the page with no error at all.

tests/orgs/test_list_query_count.py
kastger: The cost was `i.object`, not `i.get_absolute_url`. Under haystack `i` was a `SearchResult` that held only the indexed fields, so `.object` loaded the row from Postgres - one query for each row. #1825 Removed it: `i` is now the `Organization` that the page already loaded, so that line no longer queries.

PR #2782, opened 2026-08-19
#2771: Hide deprecated endpoint description format

### LINKED ISSUE #2771: API specifikacija - tik vienas pasirinkimas
Pildant paslaugos informaciją į katalogą ties "API specifikacija" katalogas leidžia pridėti tik vieną nuorodą. LRSK Teisės aktų registro DCAT'e ties viena iš paslaugų gavome keturias nuorodas šitoje DCAT skiltyje (atitikmuo DCAT formoje "Piriegos taško aprašas" / dcat: endpoint description), todėl įtraukėme tik pirmą nuorodą. Pildymas buvo vykdomas senoje pildymo formoje. 

Image

## Įgyvendinimas

- [x] "API specifikacijos formatas" lauko šalinimas
- [ ] "API specifikacija" lauko keitimas į daugiareikšmį

sirex:
## Summary

- Hide the deprecated `endpoint_description_type` field from legacy and DCAT service forms.
- Remove it from service detail pages, admin editing, validation/defaulting, RDF context, and UAPI query optimization.
- Preserve the nullable database field and historical migrations so existing values remain available internally.
- Keep endpoint URL, endpoint type, endpoint description, service-level conformance, and generated OpenAPI links unchanged.
- Remove obsolete translations and update regression tests.

## Why

`endpoint_description_type` was a Katalogas-specific extension that interpreted
`dcat:endpointDescription` as a `dcat:Distribution` with additional format metadata.
This structure is not defined by DCAT-AP/DCAT-AP-LT and was not used in practice.
The field is therefore deprecated and no longer exposed, while existing database
values are preserved for compatibility.

### DIFF
--- a/tests/datasets/test_views.py
+++ b/tests/datasets/test_views.py
@@ -397,7 +397,6 @@ class TestDatasetDetailView:
             in response.text
         )
         assert "JSON" in response.text
-        assert "OpenAPI" in response.text
         assert uapi_concept.label in response.text
 
     def test_data_service_view_without_agent(self, app: DjangoTestApp):
@@ -421,6 +420,7 @@ class TestDatasetDetailView:
         assert data_service.endpoint_url in response.text
         assert data_service.endpoint_description in response.text
         assert data_service.endpoint_type.title in response.text
+        assert endpoint_description_format.title not in response.text
 
 
 @pytest.mark.haystack
@@ -1734,13 +1734,11 @@ class TestDatasetUpdateView:
     def test_dataset_update_service_agent(self, app: DjangoTestApp) -> None:
         organization = OrganizationFactory()
         json_format = Format.objects.get(title="JSON")
-        openapi_format = Format.objects.get(title="OpenAPI")
         dataservice = DatasetServiceFactory(
             organization=organization,
             endpoint_url=None,
             endpoint_description=None,
             endpoint_type=json_format,
-            endpoint_description_type=openapi_format,
         )
         agent = AgentFactory(organization=organization)
         user = UserFactory(is_staff=True)
--- a/tests/datasets/test_views.py
+++ b/tests/datasets/test_views.py
@@ -403,12 +403,14 @@ class TestDatasetDetailView:
     def test_data_service_view_without_agent(self, app: DjangoTestApp):
         org = OrganizationFactory()
         user = UserFactory(is_staff=True, organization=org)
+        endpoint_format = Format.objects.create(title="Service endpoint format")
+        endpoint_description_format = Format.objects.create(title="Service endpoint description format")
         data_service = DatasetServiceFactory(
             organization=org,
             endpoint_url="http://test.com",
-            endpoint_type=Format.objects.first(),
+            endpoint_type=endpoint_format,
             endpoint_description="http://example.com",
-            endpoint_description_type=Format.objects.first(),
+            endpoint_description_type=endpoint_description_format,
         )
 
         app.set_user(user)
@@ -419,7 +421,6 @@ class TestDatasetDetailView:
         assert data_service.endpoint_url in response.text
         assert data_service.endpoint_description in response.text
         assert data_service.endpoint_type.title in response.text
-        assert data_service.endpoint_description_type.title in response.text
 
 
 @pytest.mark.haystack
@@ -2236,7 +2237,6 @@ class TestDatasetCreateView:
         form["contact"] = contact.pk
         form["agent"] = agent.pk
         form["endpoint_type"] = Format.objects.get(title="JSON").pk
-        form["endpoint_description_type"] = Format.objects.get(title="OpenAPI").pk
         form["conforms_to"] = Concept.objects.get(code="UAPI").pk
 
         response = form.submit()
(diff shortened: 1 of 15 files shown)


vitrina/datasets/admin.py
```diff
@@ -79,7 +79,9 @@ def get_readonly_fields(self, request, obj=None):
         if request.user.has_perm("vitrina_datasets.view_dataset") and not request.user.has_perm(
             "vitrina_datasets.change_dataset"
         ):
-            readonly_fields.extend([field.name for field in self.model._meta.fields])
+            readonly_fields.extend(
+                [field.name for field in self.model._meta.fields if field.name != "endpoint_description_type"]
+            )
```
kastger: Maybe instead of hardcoded `endpoint_description_type` value we can reuse `DatasetAdminForm` since its already imported here?:
```
excluded = set(DatasetAdminForm.Meta.exclude)
readonly_fields.extend(f.name for f in self.model._meta.fields if f.name not in excluded)
```

Note that exclude set contains more fields than that one, but I think thats still correct behaviour.

tests/datasets/test_views.py
```diff
@@ -402,12 +403,14 @@ def test_data_service_view_with_agent(self, app: DjangoTestApp):
...
+        endpoint_format = Format.objects.create(title="Service endpoint format")
+        endpoint_description_format = Format.objects.create(title="Service endpoint description format")
         data_service = DatasetServiceFactory(
             organization=org,
             endpoint_url="http://test.com",
-            endpoint_type=Format.objects.first(),
+            endpoint_type=endpoint_format,
             endpoint_description="http://example.com",
-            endpoint_description_type=Format.objects.first(),
+            endpoint_description_type=endpoint_description_format,
```
kastger, line 412: +1 on this. The PR drops the field from `ServiceResourceForm.Meta.fields` in both `vitrina/datasets/forms.py` and `vitrina/dcat/forms/dataset_forms.py`, and adds it to `DatasetAdminForm.Meta.exclude`, but `tests/datasets/test_forms.py` only removes assertions. Add a check that the field is gone from each of the three forms.

vitrina/datasets/models.py
```diff
@@ -681,6 +666,14 @@ class Dataset(Resource):
...
     # TODO: To be removed:
     # ---------------------------8<-------------------------------------
+    endpoint_description_type = models.ForeignKey(
+        "vitrina_resources.Format",
+        on_delete=models.SET_NULL,
+        verbose_name=_("API specifikacijos formatas"),
+        null=True,
+        blank=True,
+        related_name="format_endpoint_description_types",
+    )
```
kastger: Since this is in `TODO` section, it would be better if it had a follow up task related, otherwise person viewing the code won't know why this was added and why it needs to be removed.

vitrina/datasets/form_helpers.py
```diff
@@ -128,7 +123,6 @@ def validate_agent_endpoint_fields(
     endpoint_url: str | None,
     endpoint_type: str | None,
```
kastger: annotation is `str | None`, but in line 135 if reads `endpoint_type.title`. My guess is that the type should be `Format`

tests/datasets/test_views.py
kastger: This assert cannot fail since `DatasetServiceFactory` never sets `endpoint_description_type` and this PR removes the only row that rendered it. `OpenAPI` literal is in the tooltip at `coretable.html:350`, which renders for every service dataset. Worth deleting or replacing with a check on `get_endpoint_description`.

kastger: Drop `openapi_format` variable and `endpoint_description_type` argument in DatasetServiceFactory` call

tests/datasets/test_views.py
```diff
@@ -403,12 +403,14 @@ def test_data_service_view_with_agent(self, app: DjangoTestApp):
...
```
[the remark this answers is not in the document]
sirex: I decided to not add such tests, because this field will be removed eventually,  and decided to not pollute tests, with checks, that something is no longer exists.

tests/datasets/test_views.py
sirex: I was thinking about it but decided to not add so called "absence tests", because:

- **Tests bloat and maintenance** - if we will be adding "absence tests", for things we remove, such tests will accumulate and over time it will be difficult to understand, what was existed at some point, but no longer exists.

- **Integration tests** - since this field was removed or excluded from forms, this triggered number of tests fail, which I fixed, and this is a good evidence, then these fields are gone, without adding explicit absence tests.

tests/datasets/test_views.py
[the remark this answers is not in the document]
kastger: @sirex On "this field will be removed eventually" - is there a follow up task or anything that informs the next person working on this code about the planned changes?

tests/datasets/test_views.py
kastger: Regarding tests bloat and maintenance - the code is written for the situation we have now, not for planned future. If the field goes away later, deleting its tests should be part of that removal task. As of now, this plan is just an assumption, if the plan changes we then have test coverage gap.

Integration tests - I tried putting the field back on this branch in five places this PR removes it from and rand `tests/datasets/tests_forms.py`, dcat and service view tests. 300 passed, nothing failed. That matters because the field is not deleted. It stays on Dataset, so a new `ModelForm` with `fields="__all__"` puts it back on a user-facing page with green CI.

kastger (approved):

Issue #2787, opened 2026-08-22
DRAFT: Domain-driven design (DDD)


sirex:
Domain-driven design for Django

Establish vertical business boundaries and dependency direction incrementally. Keep Django as the application framework. Extract application services and domain logic only where existing workflows contain meaningful business rules. Add repositories, domain models and DTOs when they solve a demonstrated problem.

Rules:

- Views, admin, and forms do not contain business workflows.
- Application services coordinate transactions and permissions.
- Domain objects contain complex invariants that need independent testing.
- Queries may use efficient ORM projections and return DTOs.
- Repositories are introduced where a domain boundary or alternate persistence implementation is real.
- Cross-app synchronous calls use only applicaton services.
- Migrate one capability at a time, starting with a contained workflow rather than datasets or structure.

## Architectural layers

Horizontal Boundary associated with technical concerns.

1. **Presentation** - views, forms, templates, admin
2. **Application** - commands and queries
3. **Domain** - aggregates, entities, value objects, domain services
4. **Infrastructure** - models, migrations, repositories

Currently we have only two layers `Presentation` and `Infrastructure`, the application services and domain business logic is mixed somewhere between Presentation and Infrastructure.

With this proposal I recommend to add clear separation between Presentation and Infrastructure, by introducing Application and Domain layers.

Application services would be used as interface between Django apps.

Domain logic, would be responsible for reflecting data model and logic as it is defined in DCAT, UAPI and other specifications.

For example you can have two presentation implementations:

- Web UI
- UAPI

As required by EIF, to provide human and machine readable intefaces equally.

Another example for Infrastructure layer, you can have your data stored in different purpose built databases:

- PostgreSQL
- ElasticSearch (which migh be replace with PostgreSQL materialized views)
- DSA CSV files or PostgreSQL Metadata table
- OAS JSON files

Which we do this already.

By having single source of trough in the application and domain layers, we can have multiple presentations and persistance backends.

## Django apps (Bounded Context)

Vertical Boundary, a cross-cutting module for a specific business capability.

Django apps should be interpreted as Bounded Contexts of a Domain.

Each Django app should have a `domain/` directory, witch subdomains:

    {app}/
        apps.py
        urls.py              <-- Presentation: uses only Django views
        admin.py             <-- Presentation: uses only applicaton services (commands/queries)
        views.py             <-- Presentation: uses only applicaton services (commands/queries)
        templates/           <-- Presentation: uses only View objects provided by query services
[...]

PR #2790, opened 2026-08-27
#2771 part 2: The `dcat:endpointDescription` property was previous…

### LINKED ISSUE #2771: API specifikacija - tik vienas pasirinkimas

sirex:
## Summary

Part 2 of #2771. The `dcat:endpointDescription` property was previously modelled as a single `CharField` (cardinality 0..1), so a DataService could expose at most one API specification. This change raises the cardinality to 0..n by replacing the scalar field with a many-to-many relation to a new `EndpointDescription` model, allowing a data service to publish several API specifications at once.

## Changes

- **Model** (`vitrina/datasets/models.py`)
  - Added `EndpointDescription` model (UUID, `download_url` unique, URL-validated) shared across datasets via M2M.
  - `Dataset.endpoint_description` is now a `ManyToManyField` to `EndpointDescription`.
  - The old scalar column is kept as `endpoint_description_deprecated` (storage-only, `db_column="endpoint_description"`) — excluded from forms, validation, exports, detail pages and admin.
  - Added `Dataset.update_endpoint_description(urls)` which de-duplicates, shares `EndpointDescription` rows across datasets, and prunes orphans.
  - Added `Dataset.get_endpoint_descriptions()` returning sorted, de-duplicated persisted URLs plus the virtual agent-generated OpenAPI URL (merged into the same sorted set, never stored).
  - Orphan `EndpointDescription` rows are pruned on every update and on dataset deletion (`post_delete` signal).

- **Migration** `0047_endpoint_description_multivalue`
  - Creates `EndpointDescription`, renames the model field to `endpoint_description_deprecated` without touching the physical column (`SeparateDatabaseAndState`), adds the M2M, and copies legacy values into the new relation. Reversible (drops the new model data).

- **Forms** (inline `vitrina/datasets/forms.py` and wizard `vitrina/dcat/forms/dataset_forms.py`)
  - `endpoint_description` now uses `StringListField` (`unique=True`), with per-item URL validation reusing `validate_urls`. Duplicated in both forms — see "Known issues".

- **Views** — create/update flows in `DatasetCreateView`, `DatasetUpdateView`, `DcatDatasetCreateView`, `DcatDatasetUpdateView` now call `update_endpoint_description`. `DatasetHistoryView` includes `EndpointDescription` version history.

- **Exports / templates**
  - EDP DCAT-AP RDF template and `get_datasets_for_rdf` helper emit one `dcat:endpointDescription` per URL.
  - Dataset detail (`coretable.html`) lists all endpoint descriptions.

- **Admin** — `endpoint_description` shown read-only (domain logic lives in the two forms; see "Known issues"). `endpoint_description_deprecated` and `endpoint_description_type` excluded from the admin form.

- **Tests** — updated factories (`DatasetServiceFactory` uses M2M via `post_generation`), adapted existing tests, and added tests for ordering, orphan pruning, and post-delete cleanup.

- **Smoke tests** (`smoketests/`) — initial Playwright end-to-end browser tests (work in progress, see below). Added `pytest-playwright`, Chromium/Playwright to `flake.nix`, and relaxed `SESSION_COOKIE_SECURE` / `CSRF_COOKIE_SECURE` / `CSRF_TRUSTED_ORIGINS` to env-configurable for local browser testing.

## Known issues / follow-ups (intentionally not fixed here)

- **Form code duplication.** The inline and wizard forms duplicate the new domain logic. A proposed independent domain layer (#2787) was rejected for this task's scope; refactoring into one domain + two presentations is a separate task.
- **Admin read-only.** To avoid a third copy of the domain logic, `endpoint_description` is read-only in the admin.
- **Smoke tests are WIP.** Quick-and-dirty, run manually from a dev environment; they currently touch the database and a running instance rather than being fully isolated user-perspective browser tests.
- **`adms:status` confusion (#1779).** `Dataset.status` (`HAS_DATA`/`INVENTORED`/`PLANNED`/`UNASSIGNED`) is internal and not a DCAT field, yet `edp_public`/`edp_restricted` managers apply it to services. Fixed only the failing tests (`DatasetServiceFactory(status=Dataset.HAS_DATA)`); left for the dedicated issue.

## Verification

- New, updated and full test suite pass (`pytest`).
- `ruff check` / `ruff format` clean.
- New Playwright end-to-end browser test added and run manually.
- Manual testing in a web browser and manual code review; also AI-assisted code review.


vitrina/datasets/factories.py
```diff
@@ -341,6 +352,21 @@ def metadata(self, create: bool, extracted: str, **kwargs) -> None:
...
+    @factory.post_generation
+    def endpoint_description(self, create, extracted, **kwargs) -> None:
+        if not create:
+            return
+        urls = extracted
+        if urls is None:
+            return
+        if isinstance(urls, str):
+            urls = [urls]
+        from vitrina.datasets.models import EndpointDescription
```
JPriev: Is it necessary to have import inside a function? Maybe it would be better to move it in module level?

tests/api/test_views.py
```diff
@@ -3146,7 +3146,7 @@ def test_edp_dcat_ap_rdf_drops_whitespace_uris(app: DjangoTestApp):
         ),
```
JPriev, line 3146: if I understand it correctly `DatasetServiceFactory` does not have predefined `status` field as `DatasetFactory` did. For this reason `vitrina/datasets/managers.py:EdpPublicDatasetManager` will not query these datasets, because `.filter` requires `status__in=(self.model.INVENTORED, self.model.HAS_DATA)`

vitrina/datasets/factories.py
```diff
@@ -341,6 +352,21 @@ def metadata(self, create: bool, extracted: str, **kwargs) -> None:
...
 
+    @factory.post_generation
+    def endpoint_description(self, create, extracted, **kwargs) -> None:
+        if not create:
+            return
+        urls = extracted
+        if urls is None:
+            return
+        if isinstance(urls, str):
+            urls = [urls]
```
JPriev:
```suggestion
        if not create or extracted is None:
            return
        urls = [extracted] if isinstance(extracted, str) else extracted
```

JPriev (approved):

